PR #4, opened 2015-08-09
Features/osx metadata

hardkrash:
Adds ability for getting location ID from a USB serial device.


hardkrash:
Failed test on Python 2.7,  thought it worked.  Nope.

hardkrash:
I think it was being slightly premature optimization.  ```format``` and ```%``` both take more time to compute than the translate function and much less time than a ```re.replace```.  In retrospect the usage of this function is only for when ```repr``` is called.   Therefore you are right, just use format and be happy :-)

PR #23, opened 2015-10-26
There is no self.describe method, this caused an exception to be thrown

DavidHowlett:
There is no self.describe method, this caused an exception to be thrown


zsquareplusc:
thanks for finding the bug and the suggestion, but i prefer a different fix (using `description`).

DavidHowlett:
I am happy with this solution.

PR #24, opened 2015-10-28
Added a read_all function

DavidHowlett:
I work a lot with pressure sensors and flow meters that continuously stream data. Most of the time when I want to read data I want to read the entirety of what is available in the buffer.

Changing the behaviour of the serial read() method to read the whole buffer (like the file read() method reads a whole file) would probably break lots of existing code so I avoided that.


DavidHowlett:
I would update the documentation too but I am unsure how to do that.

zsquareplusc:
OK, but it's `self.in_waiting` in the new API (`inWaiting()` is only there for backwards compatibility)

And could you `git rebase` onto the current master, that way the unneeded (for the merge) commits only tracking the history would disappear (or do the change on a fresh clone when that sounds too complicated).

PR #27, opened 2015-11-02
A couple of bug fixes

DavidHowlett:
If the timeout was greater than 0 and the read function was stepped through in a debugger (or ran really slowly) then the IO could complete between the call to win32.ReadFile and win32.GetLastError. This caused an error to be thrown.

I also renamed a variable in an attempt to make the intent of the code clearer.

### DIFF
--- a/serial/serialwin32.py
+++ b/serial/serialwin32.py
@@ -273,7 +273,8 @@ class Serial(SerialBase):
                     buf = ctypes.create_string_buffer(n)
                     rc = win32.DWORD()
                     writeFinishedCorrectly = win32.ReadFile(self._port_handle, buf, n, ctypes.byref(rc), ctypes.byref(self._overlapped_read))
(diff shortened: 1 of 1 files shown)


DavidHowlett:
In the process of fixing the first bug I was tripped up by a second bug. The behaviour was:
    - if timeout == 0            then serial.read(100) returns quickly
    - if timeout == 0.001     then serial.read(100) returns quickly 
    - if 0 < timeout < 0.001 then serial.read(100) blocks
This was due to int() rounding the number of milliseconds to wait down to 0 which changed the behaviour of the windows API.

zsquareplusc:
Thanks for finding these issues. I've chosen to fix them slightly different (crediting you in the commit message).

DavidHowlett:
I am happy with your fixes. I have pulled your new master.

PR #29, opened 2015-11-03
Fixed a typo and also fixed issue with 0 timeout on windows 10

DavidHowlett:
timouts should have been timeouts

### DIFF
--- a/serial/serialwin32.py
+++ b/serial/serialwin32.py
@@ -273,7 +273,8 @@ class Serial(SerialBase):
                     buf = ctypes.create_string_buffer(n)
                     rc = win32.DWORD()
                     writeFinishedCorrectly = win32.ReadFile(self._port_handle, buf, n, ctypes.byref(rc), ctypes.byref(self._overlapped_read))
-                    if not writeFinishedCorrectly and win32.GetLastError() != win32.ERROR_IO_PENDING:
+                    err = win32.GetLastError()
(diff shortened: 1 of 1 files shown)


DavidHowlett:
In the win32 documentation for ReadFile found at:

https://msdn.microsoft.com/en-gb/library/windows/desktop/aa365467(v=vs.85).aspx

it says:

"If hFile was opened with FILE_FLAG_OVERLAPPED, the following conditions are in effect:
- The lpOverlapped parameter must point to a valid and unique OVERLAPPED structure, otherwise the function can incorrectly report that the read operation is complete.
- The lpNumberOfBytesRead parameter should be set to NULL. Use the GetOverlappedResult function to get the actual number of bytes read. If the hFile parameter is associated with an I/O completion port, you can also get the number of bytes read by calling the GetQueuedCompletionStatus function."

In the previous version of read the variable "rc" was set with ReadFile. This works on windows 7 for all the hardware that I use but on windows 10 something has changed and rc is set to 0 by ReadFile if the serial connection is to a "Beagle Bone Black". This means I am no longer able to communicate with any of my Beagle Bone Blacks. I changed the read method to use GetOverlappedResult and then it worked fine on windows 10 with all the hardware I plugged in.

bufferoverflow:
I would say it is worth to stash these changes to one commit.

DavidHowlett:
I have opened a new pull request (#50) with a single commit.

Issue #30, opened 2015-11-04
Raspberry Pi pyserial need constant reboot


AddH1008:
Hi!

I am doing an application where I need to read a RS485 stream, I tried with pyserial and the program is working once but needs reboot between each uses. Doing the same in C does not poses any problem.

The problem is more described [here](http://raspberrypi.stackexchange.com/questions/37892/raspberry-and-serial-only-working-one-shot). Apparently I am not the first to have this issue.

Can you tell something about this issue?

Thanks for the work!

AddH1008:
Sorry maybe a repost from issue #26 but I am not sure of that.

rydsfoz:
I think I have a similar problem. The application works once, if it is run a second time does not work but after 60 s application can communicate (sem precisar reiniciar o microcontrolador).

AddH1008:
To add information about it, this but does not appear at a baud rate of 2400, but is present at 115200.
@rydsfoz  if you have the possibility to switch to C code, then consider using library I linked on the stack overflow page. Good luck

MartinHjelmare:
Sounds like the same problem I've been experiencing, #26. I've downgraded to version 2.5 to avoid the problem.

AddH1008:
Hey @MartinHjelmare, interesting! Can you tell me how to get this version? I'm not that git skilled and I cannot find version earlier than the v3.0a0..

Thanks a lot

MartinHjelmare:
@AddH1008 
You should have pip installed, together with python, otherwise make sure you install pip. Then just do, from the commandline:

```
pip uninstall pyserial
pip install "pyserial==2.5"
```

pip will download from pypi and install the correct version.

AddH1008:
Genius, thanks a lot!

Issue #26, opened 2015-10-30
Reading garbage after reconnect


MartinHjelmare:
After closing the serial connection and reconnecting again, reading, using readline(), gives garbage hex.

I've tested this using python 3.4 on ubuntu 14.04 and an Arduino nano that writes to the serial port once connected. The arduino is connected to /dev/ttyUSB0. Both pyserial 2.7 and 3.0b1 show the problem, but pyserial 2.5 works nicely in this case. See console log below, for what was tested.
```
martin@martin-htpc:~/Dev/pymysensors$ mkvirtualenv -p /usr/bin/python3 pyserial
Running virtualenv with interpreter /usr/bin/python3
Using base prefix '/usr'
New python executable in pyserial/bin/python3
Also creating executable in pyserial/bin/python
Installing setuptools, pip, wheel...done.
virtualenvwrapper.user_scripts creating /home/martin/.virtualenvs/pyserial/bin/predeactivate
virtualenvwrapper.user_scripts creating /home/martin/.virtualenvs/pyserial/bin/postdeactivate
virtualenvwrapper.user_scripts creating /home/martin/.virtualenvs/pyserial/bin/preactivate
virtualenvwrapper.user_scripts creating /home/martin/.virtualenvs/pyserial/bin/postactivate
virtualenvwrapper.user_scripts creating /home/martin/.virtualenvs/pyserial/bin/get_env_details
(pyserial)martin@martin-htpc:~/Dev/pymysensors$ pip install file:///home/martin/Skrivbord/pyserial-3.0b1.zip 
Processing /home/martin/Skrivbord/pyserial-3.0b1.zip
Building wheels for collected packages: pyserial
  Running setup.py bdist_wheel for pyserial
  Stored in directory: /home/martin/.cache/pip/wheels/04/84/58/182c41f54077fd565291d425f3bb4c3213693a1a3988571de6
Successfully built pyserial
Installing collected packages: pyserial
Successfully installed pyserial-3.0b1
(pyserial)martin@martin-htpc:~/Dev/pymysensors$ pip show pyserial
---
Metadata-Version: 2.0
Name: pyserial
Version: 3.0b1
Summary: Python Serial Port Extension
Home-page: https://github.com/pyserial/pyserial
Author: Chris Liechti
Author-email: cliechti@gmx.net
License: Python
Location: /home/martin/.virtualenvs/pyserial/lib/python3.4/site-packages
Requires: 
[...]

PR #34, opened 2015-11-14
Always apply the calculated termios flags.

patthoyts:
As reported on issue #30, on Linux the use of pyserial to read serial
input from various USB serial devices seems to put the port into a state
where the second time the application is executed the data stream becomes
corrupted and unreadable.

Investigating this showed that ANY non-default termios settings solved the
problem and this patch resolves it by ALWAYS calling tcsetattr to apply
the flags calculated.

This also seems to be relevant to issue #26.

Signed-off-by: Pat Thoyts <patthoyts@users.sourceforge.net>

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -435,11 +435,10 @@ class Serial(SerialBase, PlatformSpecific):
             raise ValueError('Invalid vtime: %r' % vtime)
         cc[termios.VTIME] = vtime
         # activate settings
-        if [iflag, oflag, cflag, lflag, ispeed, ospeed, cc] != orig_attr:
-            termios.tcsetattr(
-                    self.fd,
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thanks, I did an implementation where the call to `tcsetattr` is made when opening the port but not on subsequent calls if there is no change. This should keep the original intent of he patch to not call it always, while fixing the problems in the referenced tickets.

I guess it had to do with changing timeouts, where `_reconfigure_port()` is called but it is not needed on Posix (timeouts are made with `select`), the `tcsetattr` call influenced data that was being sent or received at the same time.

patthoyts:
Excellent. I just checked your new version with the device that was demonstrating this fault and it's working fine. Thanks.

Issue #42, opened 2015-12-06
Read does not honor the timeout on Linux


cdealti:
Hello,
  Reading from the serial port with a timeout is not reliable on my system:
* Linux Mint Nadia
* Python 2.7.3
* pyserial 2.7 (but also 2.5) installed both with ```pip``` and the latest tar.gz

I have a CDC-ACM dongle that transmits data periodically.

This simple program does not respect the timeout:

```
import serial
import time

ser = serial.Serial(
                    port        = '/dev/ttyACM0',
                    baudrate    = 460800,
                    bytesize    = serial.EIGHTBITS,
                    parity      = serial.PARITY_NONE,
                    stopbits    = serial.STOPBITS_ONE,
                    timeout     = 0.1
                   )

while True:
   length = 1000000

   t0 = time.time()
   print str(len(ser.read(length))) + ' ' + str((time.time() - t0))
```

Running the above I get:

```
123 0.109111070633
246 0.205754995346
123 0.106971025467
369 0.310961008072
123 0.106863975525
369 0.309740066528
246 0.214066982269
```
Data is read but with an higher latency.
Things go worst if the timeout is set to 1s, in which case data is read with higher latencies (>> 1s).

What's going on? For comparison I've tried the following program that mimics what ```pyserial``` should do:

```
import serial
import time
import select

ser = serial.Serial(
                    port        = '/dev/ttyACM0',
                    baudrate    = 460800,
                    bytesize    = serial.EIGHTBITS,
                    parity      = serial.PARITY_NONE,
                    stopbits    = serial.STOPBITS_ONE,
                    timeout     = None, #block forever
                   )

def read(ser, length, timeout):
   n = 0
   b = bytearray()
   ser.timeout = 0 #non-blocking
   while timeout >= 0 and n < length:
       t0 = time.time()
       ready = select.select([ser], [], [], timeout)
       timeout -= (time.time() - t0)
       if len(ready[0]) != 0:
           bb = ser.read(length - n)
           n += len(bb)
           b.extend(bb)
   return b

while True:
   timeout = 0.1 #seconds
   length = 1000000

   t0 = time.time()
   b = read(ser, length, timeout)
   print str(len(b)) + ' ' + str((time.time() - t0))
```

Whose output is:

```
123 0.101263999939
123 0.101146936417
123 0.101223945618
123 0.100612163544
123 0.100650072098
123 0.101171016693
123 0.10132598877
```

As you can see data is consistently read with the specified latency.
Other runs:

[...]

PR #47, opened 2015-12-09
Fix issue #42.

cdealti:
Tested on Linux Mint Nadia with Python 2.7.3.

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -475,9 +475,11 @@ class Serial(SerialBase, PlatformSpecific):
         if not self.is_open:
             raise portNotOpenError
         read = bytearray()
+        timeout = self._timeout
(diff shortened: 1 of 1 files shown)


serial/serialposix.py
```diff
@@ -475,9 +475,11 @@ def read(self, size=1):
         if not self.is_open:
             raise portNotOpenError
         read = bytearray()
+        timeout = self._timeout
         while len(read) < size:
             try:
-                ready, _, _ = select.select([self.fd], [], [], self._timeout)
+                t0 = time.time()
```
jrobeson: can you rename this to start_time or the like?

zsquareplusc:
Thanks, it makes sense to limit the time on internal retries.

PR #64, opened 2016-01-13
utilize setuptools for scripts and package listing.

mindw:


setup.py
```diff
@@ -56,5 +53,10 @@
         'Topic :: Terminals :: Serial',
     ],
     platforms = 'any',
-    scripts = ['serial/tools/miniterm.py'],
+    entry_points={
+        'console_scripts': [
+            'list_serial_ports = serial.tools.list_ports:main',
```
zsquareplusc: why add `list_ports`? it can be called using `python -m serial.tools.list_ports`? I even considered removing miniterm from the scripts as it can be called via `-m` just as easily (left it in because of backward compatibility).

mindw: Several reasons:
1. To use `python -m serial.tools.list_ports` one must read the docs. list_serial_ports is discovered by using bash completion and be inspecting the package. I added this one after looking up the former for the nth time in the docs.
2. list_serial ports can be used even if a virtualenv is active.

mindw:
Rebased on top of 3.2.1

dpgeorge:
I'll close this because the project is moving to `pyproject.toml`; see #811.

PR #70, opened 2016-01-28
Current windows list ports does not find virtual comports. Added module to search the Windows registry for virtual comports.

HashSplat:
Find hardware comports and virtual comports in windows using both list_ports_windows and the new module list_ports_winreg.

### DIFF
--- a/serial/tools/list_ports_winreg.py
+++ b/serial/tools/list_ports_winreg.py
@@ -22,11 +22,14 @@ from serial.tools import list_ports_windows
 SERIAL_REGISTRY_PATH = 'HARDWARE\\DEVICEMAP\\SERIALCOMM'
 
 
-def regval_to_ListPortInfo(winport):
+def regval_to_listport(winport):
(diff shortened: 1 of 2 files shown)


luspock:
@justengel  Great solution! I have this problem with my virtual serial port created by com0com.  And I solved it using your method. This pull request should be accepted.

serial/tools/list_ports_winreg.py
```diff
@@ -0,0 +1,140 @@
...
+    comports = list(list_ports_windows.comports())
+
+    comports[len(comports): ] = [li for li in winreg_comports()
+                                 if li not in comports]
+
+    return comports
+# end comports_list
+
+
+def comports():
```
mfrodl: This should be `def comports(include_links=False)`.

Otherwise, the code works perfectly for me and I would also appreciate if the pull request were merged.

SeanMollet:
This should really be accepted! Pyserial on Windows is kind of broken without it.

PR #81, opened 2016-02-27
asyncio: use serial_for_url, add open_serial_connection

sbourdeauducq:
Tested OK.

### DIFF
--- a/serial/aio.py
+++ b/serial/aio.py
@@ -96,6 +96,20 @@ def create_serial_connection(loop, protocol_factory, *args, **kwargs):
     transport = SerialTransport(loop, protocol, ser)
     return (transport, protocol)
 
+@asyncio.coroutine
+def open_serial_connection(*args,
+                           loop=None, limit=asyncio.streams._DEFAULT_LIMIT,
+                           **kwds):
+    if loop is None:
+        loop = asyncio.get_event_loop()
+    reader = asyncio.StreamReader(limit=limit, loop=loop)
+    protocol = asyncio.StreamReaderProtocol(reader, loop=loop)
+    transport, _ = yield from create_serial_connection(
+        loop, lambda: protocol, *args, **kwds)
+    writer = asyncio.StreamWriter(transport, protocol, reader, loop)
+    return reader, writer
+
+
 # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 # test
 if __name__ == '__main__':
--- a/serial/aio.py
+++ b/serial/aio.py
@@ -91,7 +91,7 @@ class SerialTransport(asyncio.Transport):
 
 @asyncio.coroutine
 def create_serial_connection(loop, protocol_factory, *args, **kwargs):
-    ser = serial.Serial(*args, **kwargs)
+    ser = serial.serial_for_url(*args, **kwargs)
     protocol = protocol_factory()
     transport = SerialTransport(loop, protocol, ser)
     return (transport, protocol)


jrobeson:
can't this PR be done without the VTIME change? seems like that can be its own PR?

jrobeson:
@zsquareplusc : any comments?

jrobeson:
the VTIME change and removal of nonblocking could go to together.

serial/__init__.py
```diff
@@ -24,7 +24,7 @@
     if os.name == 'nt':  # sys.platform == 'win32':
         from serial.serialwin32 import Serial
     elif os.name == 'posix':
-        from serial.serialposix import Serial, PosixPollSerial, VTIMESerial  # noqa
```
jrobeson: is noqa no longer necessary here?

serial/aio.py
```diff
@@ -1,121 +1,127 @@
...
-#
-# This file is part of pySerial. https://github.com/pyserial/pyserial
-# (C) 2015 Chris Liechti <cliechti@gmx.net>
-#
-# SPDX-License-Identifier:    BSD-3-Clause
-"""\
-Support asyncio with serial ports. EXPERIMENTAL
-
-Posix platforms only, Python 3.4+ only.
-
```
jrobeson: lines 3 - 10 should be kept (with or without the word experimental) and likely the mention of Python 3.4+ only

serial/aio.py
```diff
@@ -1,121 +1,127 @@
...
-    def __init__(self, loop, protocol, serial_instance):
+__all__ = ["AsyncSerial"]
+
+
+class AsyncSerialBase:
+    def __init__(self, *args, loop=None, **kwargs):
+        self.ser = serial.serial_for_url(*args, **kwargs)
+
+        if loop is None:
+            loop = asyncio.get_event_loop()
```
jrobeson: Why do you prefer to start the loop here, rather than error out?

jrobeson:
the example here should be updated too https://github.com/pyserial/pyserial/blob/master/documentation/pyserial_api.rst#asyncio

serial/aio.py
```diff
@@ -1,121 +1,127 @@
...
+            return self.ser._port_handle
+
+        def read(self, n):
+            return self._loop._proactor.recv(self.fileno(), n)
+
+        def write(self, data):
+            return self._loop._proactor.send(self.fileno(), data)
+
+        def close(self):
+            self.ser.close()
```
jrobeson: these AsyncSerial classes should be renamed to something like PosixAsyncSerial and ProactorAsyncSerial

serial/aio.py
```diff
@@ -1,121 +1,127 @@
...
+        self.close()
+
+    async def read_exactly(self, n):
+        data = bytesarray()
+        while len(data) < n:
+            remaining = n - len(data)
+            data += await self.read(remaining)
+        return data
+
+    async def write_exactly(self, data):
```
jrobeson: isn't `async def` python 3.5 only?  It's still possible to do with the decorators right?

serial/aio.py
sbourdeauducq: It doesn't necessarily create a loop. https://docs.python.org/3/library/asyncio-eventloops.html#asyncio.AbstractEventLoopPolicy.get_event_loop

serial/aio.py
sbourdeauducq: Well, I replaced most of the code in this file. You may want to have a look at the latest version here: https://github.com/m-labs/asyncserial

serial/aio.py
sbourdeauducq: Yes to both.

jrobeson: so can you change it to use decorators?

serial/aio.py
jrobeson: you're right, but is it still a good idea? it says "the current policy"

serial/aio.py
sbourdeauducq: I'm not interested in < 3.5, but if you want to use the decorators instead, it's simply replacing ``async `` with ``@asyncio.coroutine`` on the line before.

serial/aio.py
jrobeson: the license header and the part about about being part of pyserial definitely needs to be kept to be accepted.

serial/aio.py
sbourdeauducq: That's what the asyncio code shipped with Python does when it has a loop parameter.

jrobeson: then I guess i'm ok with it.

serial/aio.py
jrobeson: I'd prefer not to be interested in < 3.5 myself, but it can't be helped with a common library like this one.

jrobeson:
Can you open another PR with just the async changes only? and compatible with 3.4?

zsquareplusc:
The requested features (use `serial_for_url`, add `open_serial_connection`) have been added.
A more feature complete (e.g. w/ suspend/resume) version was provided by @rob-smallshire, though it does not yet support Windows, which this version would.

serial/aio.py
sbourdeauducq: asyncserial is BSD. Feel free to integrate.

rob-smallshire:
This origin of this PR predates the work that has since been merged. In particular, take a look at 929045c9f552a88e744e51fb71ddddbc1866649f.

kurtmckee:
@zsquareplusc @dpgeorge Please close this PR. The `serial/aio.py` module has been removed long ago.

PR #90, opened 2016-03-15
Adding an USB sample

NathanDotTo:
Hi

I created a simple USB sample that you might find useful. It seems to address various questions that I found when researching this.

Regards

Nathan


jrobeson:
I think the docstrings should be added to the official documentation, and not inside this example code

PR #97, opened 2016-03-25
Makes serialposix.Serial.write() respect non-blocking

rob-smallshire:
The pySerial documentation states that a write_timeout setting of zero should result in non-blocking write() behaviour.  This pull-request modifies serialposix.Serial.write() to respect that requirement, by distinguishing between zero and non-zero numeric timeouts. In the zero timeout case, the underlying write() call is made and then the function immediately returns the number of bytes written, which may not be all of the data which which was passed to write.

The previous version of the code would iterate, using select(), until all data had been written, effectively blocking from the clients point of view.  True non-blocking writes are essential for a high-quality asyncio serial implementation.

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -521,14 +521,17 @@ class Serial(SerialBase, PlatformSpecific):
             raise portNotOpenError
         d = to_bytes(data)
         tx_len = len(d)
-        if self._write_timeout is not None and self._write_timeout > 0:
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thank you for all these nicely formatted pull requests.

rob-smallshire:
@zsquareplusc You're welcome. I have some more asyncio changes in the works to complete the details in the transport to provide async buffered writing.

PR #110, opened 2016-04-30
raise ReaderThread exceptions in connection_lost

csboling:
Default connection_lost behavior swallowed all exceptions that occurred while a ReaderThread was running, sometimes making a client protocol difficult to debug. This patch instead raises an exception if one occurred.

### DIFF
--- a/serial/threaded/__init__.py
+++ b/serial/threaded/__init__.py
@@ -55,7 +55,7 @@ class Packetizer(Protocol):
     def connection_lost(self, exc):
         """Forget transport"""
         self.transport = None
-        super().connection_lost(exc)
+        super(Packetizer, self).connection_lost(exc)
 
     def data_received(self, data):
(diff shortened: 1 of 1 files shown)


serial/threaded/__init__.py
```diff
@@ -53,6 +55,7 @@ def connection_made(self, transport):
     def connection_lost(self, exc):
         """Forget transport"""
         self.transport = None
+        super().connection_lost(exc)
```
zsquareplusc: should use old style `super` with arguments to be compatible with Python 2.7

serial/threaded/__init__.py
```diff
@@ -90,6 +93,7 @@ def connection_lost(self, exc):
         self.transport = None
         self.in_packet = False
         del self.packet[:]
+        super().connection_lost(exc)
```
zsquareplusc: should use old style `super` with arguments to be compatible with Python 2.7

PR #118, opened 2016-05-24
posix: retry if interrupted in Serial.read

nexcvon:
Retry and recalculate timeout if interrupted in Serial.read.
see https://www.python.org/dev/peps/pep-0475.


zsquareplusc:
Looks to me that `write` should also be updated and it's questionable if not every other system call needs to be done in a loop :-/

But i have not experienced problems myself and I'n not sure if there ever was an error reported regarding such probems.

PR #122, opened 2016-05-31
fix bug in FramedPacket

lfdebrux:
FramedPacket in v3.1 raises `NameError: global name 'packet' is not defined` when it receives a packet


zsquareplusc:
Thanks

PR #131, opened 2016-06-13
Only install serial.aio on Python >= 3.4.

rob-smallshire:
Changes module `serial/aio.py` into a package `serial/aio/__init__.py`.
This allows us to use `distutils` or `setuptools` to optionally install this package.  The changes to `setup.py` detect whether the current Python version is >= 3.4 and if so append `'serial.aio'` to the list of packages to be installed.

Be aware that when testing this you *must* have previously cleaned any existing distutils/setuptools build directory with:

    python setup.py clean --all

 otherwise the old aio may continue to be installed. This will only be an issue for pyserial developers.

### DIFF
--- a/serial/aio/__init__.py
+++ b/serial/aio/__init__.py
@@ -426,3 +426,4 @@ if __name__ == '__main__':
     loop.run_until_complete(coro)
     loop.run_forever()
     loop.close()
+
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,7 @@
 # (C) 2001-2016 Chris Liechti <cliechti@gmx.net>
 #
 # SPDX-License-Identifier:    BSD-3-Clause
+import sys
 
 try:
     from setuptools import setup
@@ -18,6 +19,16 @@ except ImportError:
 import serial
 version = serial.VERSION
 
+packages = [
+    'serial',
+    'serial.tools',
+    'serial.urlhandler',
+    'serial.threaded',
+]
+
+if sys.version_info >= (3, 4):
+    packages += ['serial.aio']
+
 setup(
     name="pyserial",
     description="Python Serial Port Extension",
@@ -25,7 +36,7 @@ setup(
     author="Chris Liechti",
     author_email="cliechti@gmx.net",
     url="https://github.com/pyserial/pyserial",
-    packages=['serial', 'serial.tools', 'serial.urlhandler', 'serial.threaded'],
+    packages=packages,
     license="BSD",
     long_description="""\
 Python Serial Port Extension for Win32, OSX, Linux, BSD, Jython, IronPython


rob-smallshire:
The motivation for this PR is that it should allow us not to have to use confusing Python 2 syntax in the aio module, which is Python 3.4+ only in any case.

zsquareplusc:
The "problem" with any logic for Python 2/3 in `setup.py` is, that I can no longer distribute an universal wheel, but must provide two separate files, one for each Python version. An I was hoping to reduce the work with one universal wheel ;-)

rob-smallshire:
I understand the trade-off, but given that in other PRs there are people bemoaning having to support even Python 3.4 yield syntax instead of Python 3.5 async/await I suspect that requiring they make everything work using highly obtuse Python 2.7 syntax will discourage further contributions on serial.aio.

The whole point of wheels is that we can support version-specific pre-built distributions, so I suggest we embrace that approach, and write a short script to generate and upload the requisite wheels. I would volunteer to do so, but I'm not sure exactly how you approach distribution at your end.

zsquareplusc:
Hm, two wheels would probably not work to support the same versions as today... pySerial currently runs on Python >=3.2. With a feature only working on Python 3.4 onward, it would require separate wheels for every version, as I understand it, the specification only allows all of Python (universal) the all of 2.x respectively 3.x or single versions, but not ranges of versions (even when it did, it would still require 3 packages).

The other option would be two make a second package. That way, pySerial could stay "universal" and the Python 3.4 (or even 3.5) features would be in a separate package, e.g. `pyserial-asyncio`. It would also make sense regarding "API stability" and supported OS. I tend to this solution.

rob-smallshire:
I'm very much inclined to agree that extracting the pyserial.aio into a separate package is the way to go, especially as aio is currently strictly layered on top of the core pyserial code. It's certainly preferable solution to having us write Python 3.4 code in Python 2!

zsquareplusc:
The tricky part is then to choose a good name for the package (I think `pyserial-asyncio` would be good as a distribution name). I do not think that a second package can install "into" the `serial` namespace.

So either pySerial would need an import redirection, which is probably not a good idea. Or the new package has a different name. Maybe `serial_asyncio`.

rob-smallshire:
I agree `pyserial-asyncio` is the right PyPI package name. I believe it *should* be possible to make it appear as `serial.aio` using Python's little used 'namespace packages' feature. I'll read up on this tomorrow and do some experiments.

zsquareplusc:
Seems Python 3.3 and newer should support that by [PEP-420](https://www.python.org/dev/peps/pep-0420/). From skimming through the document, I'm not sure if this will work for pySerial. A key to make a namespace package is to leave out `__init__.py`. This can, however, not be removed to keep compatibility with Python 2. I'm not sure if leaving away `__init__.py` from the sub package only, is sufficient. The PEP is not specific in that detail.

rob-smallshire:
PEP 420 is about _implicit_ namespace packages. We can use regular namespace packages with `__init__.py` files by using the Python Standard Library `pkgutil` module at least as far back as Python 2.7 

Take a look at #134 which explores this approach. I think it's the way to go.

zsquareplusc:
So i'll reject this one and accept #134. Do you want to transfer your `pyserial-asyncio` repo to the pyserial "organization"? (maybe you need to do that indirectly via me, as according to the [docs](https://help.github.com/articles/transferring-a-repository-owned-by-your-personal-account/) user to  organization has more restrictions).

rob-smallshire:
In principle, yes, I'd think it's good to transfer ownership to your 'pyserial' organisation. In practice, that looks quite awkward though. I think it's easiest if you:

1. Clone my `pyserial-asyncio` repository locally:

    git clone https://github.com/rob-smallshire/pyserial-asyncio.git

2. Create a new, *completely empty* `pyserial/pyserial-asyncio` GitHub repository owned by your organisation.

3. Update the remote URL for your local clone:

    git remote set-url origin https://github.com/pyserial/pyserial-asyncio.git

4. Push everything to your new GitHub repo:

    git push origin master

Let me know when you're done, and I'll delete my GitHub repo to avoid confusion.

zsquareplusc:
I think transferring to me would work (and i would transfer ti to pySerial). That way you are registered in the git history correctly.

rob-smallshire:
Ah, that was easier than expected. It's on its way.

PR #140, opened 2016-07-08
Added documentation for is_open and isOpen.

MatthewWest:
I'm not sure if isOpen() has been officially deprecated, so I just marked it as such.


zsquareplusc:
Yes that is fine, thanks.

PR #141, opened 2016-07-27
linux: list_ports use interface path for multi location

eliotb:
On devices with multiple serial interfaces, like FT4232,
the usb_device path is the same for all interfaces, so hwid does
not uniquely identify the interface.
So in this case use the basename of the interface directory.

Single interface:
hwid: USB VID:PID=0403:6015 SER=DN00RHW0 LOCATION=3-1.1
Multiple interface:
hwid: USB VID:PID=0403:6011 SER=FT97WK0I LOCATION=3-1.2:1.1

Signed-off-by: Eliot Blennerhassett eliot@blennerhassett.gen.nz

### DIFF
--- a/serial/tools/list_ports_linux.py
+++ b/serial/tools/list_ports_linux.py
@@ -28,17 +28,28 @@ class SysFS(list_ports_common.ListPortInfo):
             self.subsystem = None
         # check device type
         if self.subsystem == 'usb-serial':
-            self.usb_device_path = os.path.dirname(os.path.dirname(self.device_path))
+            self.usb_interface_path = os.path.dirname(self.device_path)
         elif self.subsystem == 'usb':
-            self.usb_device_path = os.path.dirname(self.device_path)
+            self.usb_interface_path = self.device_path
         else:
-            self.usb_device_path = None
+            self.usb_interface_path = None
         # fill-in info for USB devices
-        if self.usb_device_path is not None:
+        if self.usb_interface_path is not None:
+            self.usb_device_path = os.path.dirname(self.usb_interface_path)
+
+            try:
+                num_if = int(self.read_line(self.usb_device_path, 'bNumInterfaces'))
+            except ValueError:
+                num_if = 1
+
             self.vid = int(self.read_line(self.usb_device_path, 'idVendor'), 16)
             self.pid = int(self.read_line(self.usb_device_path, 'idProduct'), 16)
             self.serial_number = self.read_line(self.usb_device_path, 'serial')
-            self.location = os.path.basename(self.usb_device_path)
+            if num_if > 1:  # multi interface devices like FT4232
+                self.location = os.path.basename(self.usb_interface_path)
+            else:
+                self.location = os.path.basename(self.usb_device_path)
+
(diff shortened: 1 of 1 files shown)


eliotb:
To clarify with a fuller example.

/dev/ttyUSB9 is on a device with only one interface.  This is what the output looks like without this patch.

/dev/ttyUSB10..13 are all ports on the same FT4232 device.  Their hwids are now distinct, with added 1.0, 1.1, 1.2, 1.3. Without the patch, the four hwids are identical.

/dev/ttyUSB9        
    desc: VScom USB-COM Plus - Config
    hwid: USB VID:PID=0403:6015 SER=DN00RHW0 LOCATION=3-1.1
/dev/ttyUSB10       
    desc: VScom USB-COM Plus
    hwid: USB VID:PID=0403:6011 SER=FT97WK0I LOCATION=3-1.2:1.0
/dev/ttyUSB11       
    desc: VScom USB-COM Plus
    hwid: USB VID:PID=0403:6011 SER=FT97WK0I LOCATION=3-1.2:1.1
/dev/ttyUSB12       
    desc: VScom USB-COM Plus
    hwid: USB VID:PID=0403:6011 SER=FT97WK0I LOCATION=3-1.2:1.2
/dev/ttyUSB13       
    desc: VScom USB-COM Plus
    hwid: USB VID:PID=0403:6011 SER=FT97WK0I LOCATION=3-1.2:1.3

zsquareplusc:
Seems to be useful.

But the same feature should be added for win32 and OSX too (can also be in a separate patch / later / by someone else). Do you know why the interface is a two digit number and not just one? If the location is replicated on the other two OS, it should probably look similar.

eliotb:
AFAIK Windows hwid string doesn't include location at all (just vid:pid, ser)

For linux I just followed the existing pattern using the final element in the device path.
I'm not sure why they are 1.1, 1.2 rather than just 1, 2,...
Maybe the 1. is the configuration number.

A full path looks like:
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-1/3-1.2/3-1.2:1.0

and the corresponding fragment of the USB topology:
/:  Bus 03.Port 1: Dev 1, Class=root_hub, Driver=xhci_hcd/4p, 480M
    |__ Port 1: Dev 82, If 0, Class=Hub, Driver=hub/4p, 480M
        |__ Port 2: Dev 84, If 0, Class=Vendor Specific Class, Driver=ftdi_sio, 480M

zsquareplusc:
I added a [windows](https://github.com/pyserial/pyserial/blob/master/serial/tools/list_ports_windows.py#L236) and [OSX](https://github.com/pyserial/pyserial/blob/master/serial/tools/list_ports_osx.py#L186) implementation for the LOCATION a while ago.

jrmrjnck:
Hey, I just submitted an issue about the same thing: https://github.com/pyserial/pyserial/issues/168. It would be great to get this merged.

The x.y after the colon is indeed configuration.interface.

Like I noted in my issue, the interface number can be obtained on Windows from the HW ID string. It is not included in the location string, which only goes down to the USB device.

PR #144, opened 2016-08-03
fix encoding error on windows

jat001:
Some attributes may have non-english characters (like ```info.description```) and are not utf-8 encoded.
```locale.getpreferredencoding``` will return ANSI code page (on my system, it is ```cp936```) so we can re-decode the strings with it.


zsquareplusc:
I see, but re-encoding to a byte stream should not be needed, it can be an unicode string

jat001:
Indeed, it is an unicode string but with wrong encoding. It is necessary to restore it as a sequence of bytes.

zsquareplusc:
I have implemented a different fix. Using the Windows Unicode API (`*W` instead of `*A` functions) its not needed to convert the encoding.

PR #149, opened 2016-08-12
Fix "TypeError: an integer is required" error from data_received.

f1sty:
Change append to extend in serial.threading.FramedPacket data_receive function, wich will work both under Python 2.7, and Python 3.5, without throwing and error.


zsquareplusc:
surprising, at first, that for appending a single item, `extend`works better than `append`...
but for Python 2.x  a single character or byte is the same as a sequence of one item, so both methods work.

Thanks for the fix.

f1sty:
Nice. Thanks for the merge and feedback.

PR #161, opened 2016-09-12
spelling

DavidHowlett:


zsquareplusc:
Thanks, I've fixed it in  a commit of mine and mentioned you instead of using this pull request as there were a bit many commits to just remove a character ;-)

PR #164, opened 2016-09-23
Update serialwin32.py

nick83ola:
Missing a semicolon!!!


zsquareplusc:
Thanks, sorry, missed this patch and fixed it in an other commit.

PR #169, opened 2016-10-07
Add BAUDRATE_LITERAL flag, for the case of *BSDs where arbitrary baud…

kevans91:
… rates are generally accepted

This is at the very least true for [NetBSD](http://netbsd.gw.com/cgi-bin/man-cgi?tcsetattr+3+NetBSD-current), [FreeBSD](https://www.freebsd.org/cgi/man.cgi?tcsetattr), and [OpenBSD](http://man.openbsd.org/OpenBSD-6.0/tcsetattr.3) and likely covers any other *BSD you may come across, so we just do a catch-all (with explicit for these three cases) and can sort things out later if the fail or are reported.

This is implemented as a fallback alternative to the BAUDRATE_CONSTANTS because it does not need anything special for setting -- just set ispeed/ospeed to the baudrate and go about your business. Will it fail later? Possibly, but that's fairly rare.

This was tested on FreeBSD -CURRENT, but I've been doing this in non-Python land for a bit (still using termios) with Arduino-related stuff. =)

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -43,6 +43,7 @@ from serial.serialutil import SerialBase, SerialException, to_bytes, \
 
 class PlatformSpecificBase(object):
     BAUDRATE_CONSTANTS = {}
+    BAUDRATE_LITERAL = False
 
     def _set_special_baudrate(self, baudrate):
         raise NotImplementedError('non-standard baudrates are not supported on this platform')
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thanks, but I think there is a solution without adding a new concept (`BAUDRATE_LITERAL`). See my commit referenced above.

kevans91:
Ah, great! Thank you. =) Definitely cleaner.

PR #172, opened 2016-10-28
Windows: list_ports with 'manufacturer' info property

rabryan:
On windows when reading port info, get the 'manufacturer' property similar to other platforms.


zsquareplusc:
Cool, thanks

PR #189, opened 2016-12-13
Fix the return value of write()

fahhem:
It's documented as returning the number of bytes written, but `d` gets sliced as it goes along.


zsquareplusc:
Thank you.

But I think your fix was incomplete (taking `len(orig_len)`  instead of the value itself). Did my own implementation.

PR #190, opened 2016-12-20
Fix typos in documentation

jwilk:
Found using [mwic](http://jwilk.net/software/mwic) and [anorack](https://github.com/jwilk/anorack).


zsquareplusc:
Cool, thank you.

PR #206, opened 2017-02-24
Exclusive access on POSIX.

NJDFan:
On Linux machines, opening a serial port is non-exclusive; you can open multiple simultaneous connections to the same serial port.  This does not seem to be the case on Windows, but more the point is almost certainly not the behavior you want, various processes all stepping on one another's comm.

Unfortunately you can't just open the Serial object then fcntl.flock it because, in between, initializing the Serial port did all manner of things to settings like baud rate, timeouts, etc.  So the fix is that the call to flock has to happen immediately, right after the os.open of the raw port, before any of the termios action.

This pull adds an exclusive keyword argument to Serial.__init__ and corresponding data descriptors.  On a serialposix.Serial, setting exclusive explicitly True or False manipulates the lock for this process, meaning that if you make two attempts to open a Serial(port, exclusive=True), the second will fail.  On serialwin32 and serialjava, the exclusive flag is supported but ignored.  Windows already does the right thing, and I don't have access to a Jython installation to check the Java behavior.

This change is kind of essential for applications that try to do autodetection of "Alright, which serial port has my Widgetek brand Widget."

### DIFF
--- a/documentation/pyserial_api.rst
+++ b/documentation/pyserial_api.rst
@@ -12,7 +12,7 @@ Native ports
 
 .. class:: Serial
 
-    .. method:: __init__(port=None, baudrate=9600, bytesize=EIGHTBITS, parity=PARITY_NONE, stopbits=STOPBITS_ONE, timeout=None, xonxoff=False, rtscts=False, write_timeout=None, dsrdtr=False, inter_byte_timeout=None)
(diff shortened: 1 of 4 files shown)


zsquareplusc:
Thanks, this looks useful.

Maybe the other platforms should override the `exclusive` property, e.g. on Windows to only accept/return `True' . I'll look into this.

PR #208, opened 2017-03-09
Extend PTY test case

lkarsten:
Send bytes through the virtual port and verify that they arrive.

Background: Used to rule out why serial.read() was blocking, but fd.read()
from fdopen() was not.


zsquareplusc:
Good idea but implementation failed on Python 3 (`bytes`). I've committed an extended test.

PR #211, opened 2017-03-19
changed modem line status methods to setters and getters

gesorthy:
Modem line status methods (setRTS, setCTS, getRI, ....) are deprectaed in PySerial 3, although they are still present in rfc2217.py. Replaced methods by setters and getters.

### DIFF
--- a/serial/rfc2217.py
+++ b/serial/rfc2217.py
@@ -613,9 +613,7 @@ class Serial(SerialBase):
             while len(data) < size:
                 if self._thread is None:
                     raise SerialException('connection failed (reader thread died)')
-                data += self._read_buffer.get(True, timeout.time_left())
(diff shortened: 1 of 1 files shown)


serial/rfc2217.py
```diff
@@ -613,9 +613,7 @@ def read(self, size=1):
             while len(data) < size:
                 if self._thread is None:
                     raise SerialException('connection failed (reader thread died)')
-                data += self._read_buffer.get(True, timeout.time_left())
-                if timeout.expired():
-                    break
+                data += self._read_buffer.get(True, self._timeout)
```
zsquareplusc: unfortunately, this is not equivalent. The original code uses a special timeout object (that uses a monotonic clock if available).

PR #213, opened 2017-03-23
Changed modem line status methods to setters and getters.

gesorthy:
Although modem line status methods (like getRTC, getRI, ...) are deprecated in Pyserial 3, they are still present in rfc2217.py. Replaced methods by corresponding properties.


zsquareplusc:
Thanks.

FYI, you can also update a merge request, no need to create a new one.

PR #225, opened 2017-04-24
protocol_socket: Retry if BlockingIOError occurs in reset_input_buffer.

mawesi:
On some windows machines opening a RFC 2217 connection results in an **BlockingIOError: [WinError 10035]**.  As this error is nonfatal and the operation should be retried later according to [Windows Sockets Error Codes](https://msdn.microsoft.com/en-us/library/windows/desktop/ms740668%28v=vs.85%29.aspx) simply ignoring this exception solves the problem.

### DIFF
--- a/serial/urlhandler/protocol_socket.py
+++ b/serial/urlhandler/protocol_socket.py
@@ -239,6 +239,9 @@ class Serial(SerialBase):
             ready, _, _ = select.select([self._socket], [], [], 0)
             try:
                 self._socket.recv(4096)
+            except BlockingIOError:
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thank you

wolfmanjm:
This also needs to be done in posixserial.py as I get BlockingIOError with async-serial

zsquareplusc:
ah, silly... i posted the thank you message last time, but forgot to press the merge button (or the confirmation) :/

zsquareplusc:
I changed the fix, with the commit mentioned above, so that it is backwards compatible with Python 2.7

PR #227, opened 2017-05-03
Handle EAGAIN on write (BlockingIOError: [Errno 11] Resource temporar…

wolfmanjm:
…ily unavailable)

As the pyserial-asyncio uses non blocking mode posix serial needs to handle the EAGAIN error which is a non fatal error that is often seen in non blocking mode.

Without it we get this error in asyncio mode when writing a lot of data..

```
protocol: <__main__.SerialConnection object at 0xb6d0bd8c>
transport: SerialTransport(<_UnixSelectorEventLoop running=True closed=False debug=False>, <__main__.SerialConnection object at 0xb6d0bd8c>, Serial<id=0xb6d0b9ac, open=True>(port='/dev/ttyACM0', baudrate=115200, bytesize=8, parity='N', stopbits=1, timeout=0, xonxoff=False, rtscts=False, dsrdtr=False))
  File "/usr/local/lib/python3.4/dist-packages/pyserial-3.3-py3.4.egg/serial/serialposix.py", line 534, in write
    n = os.write(self.fd, d)
BlockingIOError: [Errno 11] Resource temporarily unavailable
```

Fixes this issue https://github.com/pyserial/pyserial-asyncio/issues/20

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -560,6 +560,9 @@ class Serial(SerialBase, PlatformSpecific):
                 tx_len -= n
             except SerialException:
                 raise
+            except BlockingIOError:
+                # EAGAIN is common when in non blocking mode
+                pass
             except OSError as v:
                 if v.errno != errno.EAGAIN:
                     raise SerialException('write failed: {}'.format(v))


zsquareplusc:
OK but the fix only works for Python 3, that Exception name is not present in Python 2.7 (I need to review #225 too).

wolfmanjm:
oh I am sorry I forgot to test on 2.7 as asyncio only works on 3... I'll see what I can do for 2.7.
thank you

zsquareplusc:
I've ported the code from the socket handler here and saw that EINTR was not handled for Py 2.7 while it was for 3.x.

So you do not need to provide an updated patch, but thanks for finding the issue.

wolfmanjm:
@zsquareplusc Thank you, however that fix is for read. The error I get is with write not the read.
So this still needs to be added to the write method as well as my patch does.

zsquareplusc:
OK, i think it makes sense to catch the same errors in read and write, pushed a change with that.

wolfmanjm:
Hi @zsquareplusc I just tested your fix.. however I now get this error... Seems that it still is not handling non blocking mode correctly on writes.
I'll try to debug further. My original fix did not get this error.

'''
protocol: <__main__.SerialConnection object at 0xb6c8c40c>
transport: SerialTransport(<_UnixSelectorEventLoop running=True closed=False debug=False>, <__main__.SerialConnection object at 0xb6c8c40c>, Serial<id=0xb718fb0c, open=True>(port='/dev/ttyACM0', baudrate=115200, bytesize=8, parity='N', stopbits=1, timeout=0, xonxoff=False, rtscts=False, dsrdtr=False))
  File "/home/morris/.local/lib/python3.4/site-packages/pyserial_asyncio-0.4-py3.4.egg/serial_asyncio/__init__.py", line 120, in write
    n = self._serial.write(data)
  File "/home/morris/.local/lib/python3.4/site-packages/pyserial-3.3-py3.4.egg/serial/serialposix.py", line 577, in write
    raise writeTimeoutError
serial.serialutil.SerialTimeoutException: Write timeout
'''

wolfmanjm:
ok the issue appears to be here... https://github.com/pyserial/pyserial/blob/10d79d9dcd508a391162f1c1b68538e60e5e931e/serial/serialutil.py#L135

even though the Timeout is set to 0 ie non blocking this https://github.com/pyserial/pyserial/blob/10d79d9dcd508a391162f1c1b68538e60e5e931e/serial/serialutil.py#L129 sets the target_time when it should be set to None. this causes a bogus timeout error for non blocking writes.

wolfmanjm:
Changing 
https://github.com/pyserial/pyserial/blob/10d79d9dcd508a391162f1c1b68538e60e5e931e/serial/serialutil.py#L147
to
` if duration is not None and duration != 0:` 
seems to fix this issue, but I am not convinced that is correct, what do you think?

wolfmanjm:
Possibly the correct fix is to do what was there originally and not call in https://github.com/pyserial/pyserial/blob/10d79d9dcd508a391162f1c1b68538e60e5e931e/serial/serialposix.py#L576
```
           if timeout.expired():
                raise writeTimeoutError
```
unless there is an exception.

zsquareplusc:
It was probably broken before that change too... (your fix had a separate `except` clause that skipped the test). It seems that exceptions in non-blocking I/O would have triggered the faulty write error.

The timeout object returns `True` for `expired()` if the timeout is 0 (non-blocking I/O). This is fine for `read()` where it just breaks out of the read loop, but not good for `write()` where it would throw an exception (the return for non-blocking I/O is separate here). The fix is to skip the test for non-blocking I/O.

wolfmanjm:
SKip the test for non blocking write in posixserial or in Timeout? I am not clear where to exactly fix this :)

wolfmanjm:
Ahh ok never mind I see your fix, thank you I'll test it ASAP.

wolfmanjm:
Thank you...

I tested the latest and it works perfectly.

Issue #168, opened 2016-10-03
Request: Expose USB interface descriptor index


jrmrjnck:
I am trying to connect with a USB device which exposes two CDC/ACM functions (not sure if function is the right word - I'm new to USB terminology). There are a total of four USB interfaces (control+data for two functions). In pyserial, the two functions are detected but they have identical ListPortInfo data, so it's impossible to know which is which.

I took a look at list_ports and it should be possible to get the interface information (at least on Windows and Linux, which is what I need to support).

On Windows, the interface index shows up as `&MI_XX` at the end of the HW ID string. (e.g. `USB\VID_1FC9&PID_0003&REV_0100&MI_02` for the second function on interface index 2).

On Linux, the sysfs path includes the config.interface information following a colon `:`.

diamondman:
@jrmrjnck Looks like your request might be done. Please check, and see if we can get this issue closed :).

jrmrjnck:
I no longer use the hardware which prompted this issue report, so I can't verify it. But since it's not relevant for me any longer, I'm happy to close it.

PR #230, opened 2017-05-23
Added query also for &MI_xx parameter from HardwareID string

ciorceri:
Some USB Composite Devices will have a MI ('multiple interfaces')
parameter which indicates what is the interface number of that USB
device.
If this &MI_xx parameter is present the current regexp will not work correctly and it will not get the 'serial_number'. This will fix also that issue.


zsquareplusc:
Is related to #168 and #141

zsquareplusc:
merging, but i'll later do some refactoring to combine it with #141

PR #231, opened 2017-05-24
Stopped self being passed to time.time and time.monotonic

JamesStewy:
This PR stops `self` being passed as an argument to `time.time` and `time.monotonic`. This behaviour caused an issue when overriding `time.monotonic` with the `monotonic` package in python 2.7 ( https://github.com/UAVCAN/pyuavcan/issues/22).

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -113,12 +113,14 @@ class Timeout(object):
         # supported by Python 3.3 and above. It returns a time in seconds
         # (float) just as time.time(), but is not affected by system clock
         # adjustments.
-        TIME = time.monotonic
+        def TIME(self):
+            return time.monotonic()
     else:
         # Timeout implementation with time.time(). This is compatible with all
         # Python versions but has issues if the clock is adjusted while the
         # timeout is running.
-        TIME = time.time
+        def TIME(self):
+            return time.time()
 
     def __init__(self, duration):
         """Initialize a timeout with given duration"""


chrismurf:
Recently ran into the same issue with my own monkeypatched version of monotonic - support merging.

dpgeorge (approved):
Thanks for the patch.  The way the code is currently written, it's definitely problematic.

Although the change here is a little inefficient (makes a new wrapper function), I think it's the best solution because it's compatible with the `test/test_timeout_class.py` test, and doesn't break an API in case users were using `Timeout.TIME` for some reason.

zsquareplusc:
I'm not sure if that fix is (still) needed?

storing references to callables may it be a function or bound method should not be a problem in general. I think the problem here is that the `TIME` attribute is set at class creation and then there is a mismatch if the time module is monkey patched later.

maybe nowadays we could just directly use `time.monotonic` (in place of `self.TIME`) and drop the fallback to `time.time`, as all the modern Python since at least 3.5 have that.

dpgeorge:
> storing references to callables may it be a function or bound method should not be a problem in general. I think the problem here is that the `TIME` attribute is set at class creation and then there is a mismatch if the time module is monkey patched later.

No, I think the issue is that if the `time` module is mocked, or `monotonic` is otherwise implemented in Python (and not a built-in) then using `time.monotonic` as a method of a class will pass self into that function as the first argument.

It's a bit strange, but Python treats built-in functions differently to functions implemented in Python, when used as a method.

Eg:
```py
import time

time.monotonic = lambda: 42       

class Timeout:
    TIME = time.monotonic

    def get_time(self):
        return self.TIME()

print(Timeout().get_time())
```
That will give an error `TypeError: <lambda>() takes 0 positional arguments but 1 was given`, because it tries to pass `self` to the overridden `monotonic` function.  But if you comment out the `time.monotonic = ` line then it works!

So, IMO this PR is needed to fix that problem, in the case that `time.monotonic` is not a built-in C function.


> maybe nowadays we could just directly use `time.monotonic` (in place of `self.TIME`) and drop the fallback to `time.time`, as all the modern Python since at least 3.5 have that

Considering that Python 3.10 or newer is now required (since 388c8b5eec2cd0435fa91b3778c4e796045fc096) it would definitely make sense to just use `time.monotonic` everywhere.  Or even better, use `time.monotonic_ns` because that won't lose precision for large timestamps (and it's available since Python 3.7).

But then the unit test `test/test_timeout_class.py` needs updating, and users won't be able to subclass `Timeout` to change the `TIME` entry (like that test does).  Although I doubt anyone does that in practice.

kurtmckee:
This PR can be closed. Python 2.7 is no longer supported, and `time.monotonic()` exists on all supported versions of Python.

dpgeorge:
> This PR can be closed. Python 2.7 is no longer supported

This is not just about Python 2.7, the same issue occurs with Python 3 if you attempt to override/mock the `time.monotonic` function (eg for a unit test).

IMO this PR should be merged, due to Python's method binding rules.

kurtmckee:
A change is needed, but this PR is not that needed change. Please see #842, which supersedes this PR and demonstrates mocking `time.monotonic()` as part of testing. :+1:

dpgeorge:
Closing in favour of #842.

PR #240, opened 2017-06-28
serial: SerialBase with is idempotent

AmberAussie:
Hi,

I got the issue when I was doing something like:

    import serial

    port = serial.Serial(
        port='COM4',
        baudrate=115200,
        bytesize=8,
        parity=serial.PARITY_NONE,
        stopbits=1,
        timeout=2)

    with port as stream:
        stream.write(b'1234')

    # do other stuff like calling an exe which use COM4

    with port as stream:
        stream.write(b'5678')

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -557,6 +557,8 @@ class SerialBase(io.RawIOBase):
     # context manager
 
     def __enter__(self):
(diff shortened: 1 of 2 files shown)


zsquareplusc:
Thanks

PR #250, opened 2017-07-27
Don't reset input buffer on open

### LINKED ISSUE #249: Early data missed on open on Posix systems
serialposix.Serial.open calls self.reset_input_buffer() after opening which causes the underlying data that's been buffering to be dropped.  FWIW serialwin32.Serial.open does not call this function on open.

vlovich:
Fixes #249 which causes data that has been buffering in the OS to be dropped on Posix systems with no way to recover.  Windows does not reset_input_buffer on open.

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -292,7 +292,6 @@ class Serial(SerialBase, PlatformSpecific):
                 pass
             else:
                 raise
-        self.reset_input_buffer()
         self.pipe_abort_read_r, self.pipe_abort_read_w = os.pipe()
         self.pipe_abort_write_r, self.pipe_abort_write_w = os.pipe()
         fcntl.fcntl(self.pipe_abort_read_r, fcntl.F_SETFL, os.O_NONBLOCK)


nosredna:
I support this fix and I believe flushing the input should be left up to the application.

flowswitch:
I vote for this too. A real case: flush drops the connection setup packet from Qualcomm Sahara bootloader.
Also looks like the serialjava.py open doesn't flush.

dpgeorge:
I also support this change / fix (I guess it's a fix because the Windows backend does not flush on open).

But it would be a change in behaviour, so needs input from @zsquareplusc .

Note that this PR itself needs reworking on latest master due to f28543bee0b132b2977848159b4690bca86b05f0, but the change is still logically the same, just remove the call to `self._reset_input_buffer()`.

zsquareplusc:
probably OK to change to not reset the input on open.

the original idea was to have a clean start, which is fine for many use cases. RTS/DTR is also switched on open so devices could have some reaction when the port is opened. sometimes one could get a 0xff or data from a previous session. but then again, good applications are robust even is some faulty data is received and the clean startup can be achieved by calling `reset_input_buffer()` in an application.

kurtmckee:
PR #829 merges `master` into this PR's work to resolve merge conflicts.

@zsquareplusc @dpgeorge Please close this PR and merge #829.

PR #257, opened 2017-08-02
Update README.rst

shaunwbell:
Add conda install instructions


zsquareplusc:
OK, merging. But i think an older pySerial is also installed by default in conda. M;ybe that could or should be mentioned?

shaunwbell:
Looks like a comment on the readthedocs source would be a good spot for this.

addressed in #259

Issue #262, opened 2017-08-20
macOS ListPortInfo object short name is None


BrendanSimon:
On macOS, the `ListPortInfo` object returned by `list_ports.comports()` has the `name` attribute set to `None`.

I have to workaround this using the `device` attribute and convert to a short name using `os.path.basename(device)`

PySerial 3.4
Python 3.6.2
macOS Sierra 10.12.6

BrendanSimon:
Could this be fixed by placing the following at line 36 of `lists_ports_common.py`?

`self.name = os.path.basename(device)`

and can remove the equivalent line at line 27 of `lists_ports_linux.py`

That way the short name is set for all ports, at least as a default and can always be overridden for a specific platform if desired.

zsquareplusc:
fixed in #263

PR #263, opened 2017-08-20
list_ports: set default `name` attribute

BrendanSimon:
See #262

Set the default `ListPortInfo.name` attribute from the full `device` name,
using `os.path.basename(device)`

This should be valid for all platforms, but can be easily overridden in
platform specific code that populates the `ListPortInfo` object.

### DIFF
--- a/serial/tools/list_ports_common.py
+++ b/serial/tools/list_ports_common.py
@@ -10,6 +10,7 @@
 import re
 import glob
 import os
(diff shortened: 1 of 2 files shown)


zsquareplusc:
OK, lets hope that does not mask other errors. I'll also remove the `None` default in `__init__` as that would not make sense with this patch.

Issue #234, opened 2017-06-06
[miniterm.py] Add option to save output to a file?


marrakchino:
Hi,
I'm here to enquire about any plans to add a parameter which enables copying serial output to a file (sort of a log file to keep a trace of the serial communication) to **miniterm**.

Regards.

ssldemon:
`miniterm.py @port @baudrate > output.txt`
should do the trick. my output is directed to output.txt when I use that

nelqatib:
Does it preserve printing to standard output? I'm not sure...
The enhancement I asked about was to have both real-time printing to standard output, and (in the background), appending the same output to a file.

nelqatib:
@zsquareplusc would you be interested if I work on this feature?

craighurley:
On *nix based systems, if you want to view the serial traffic in the terminal as well as log to a file:

```sh
miniterm.py @port @baudrate | tee output.log
```

PR #264, opened 2017-08-21
miniterm: Implement log file feature

n-eq:
Hi,

This commit implements the `--log` option (closes https://github.com/pyserial/pyserial/issues/234) to append the serial output
to a log file (in parallel with terminal display). It allows the user to
save the serial output of a session for further operations (compare two
outputs for example).

Usage:
`miniterm.py @port @baudrate --log LOG_FILE`

Please tell me if any documentation file should be changed.

Regards


nelqatib:
I tried to cherry-pick the important commit so I had to rewrite history. I may have messed it up.
If needed, I can open a new PR.

StinGer-ShoGuN:
I've not gone through all the commits listed here, but from what I understand, you just open a log file every time and write to it the transformed text.

While I have actually nothing against (simple, light, efficient, great !), I've done another implementation of this feature using a filter.
Benefits are as follow:
- enable/disable log in a session
- change log file in a session
- have multiple log files,
- choose when/what to log (before or after this or that transformation).

I've also done a filter to prepend timestamps on new lines and the log filter I've done allow me to log the session without those timestamps if needed.

I could make a PR for this log feature of mine and let the devs here choose which one to use.

nelqatib:
@StinGer-ShoGuN Yes, seems great!
The additional features you suggest are really nice. Indeed, my work implements the log file feature with out the possibility of chosing/closing one within a session, etc..
It seems you commits are much more accomplished. 
Could you please specify how to use your feature?

StinGer-ShoGuN:
You can use it in two ways:
- add a `-f log` to the command line when starting `miniterm` -- you can specify it multiple times so that you will have multiple log files;
- press `ctrl + t ctrl + f` within a session and add a `log` filter (or again, several).

You will then be prompted for log file name to log to.

Examples
========
WARNING: all of those only work with my repo !

##### Example 1
```sh
python -m serial.tools.miniterm --eol CRLF -f log -f default /dev/ttyUSB1 1000000
```
Log everything after `default` filter has been applied on RX data and before on TX data (this is how I understood it was made to be in `miniterm`).

##### Example 2
```sh
python -m serial.tools.miniterm --eol CRLF -f log -f default -f log /dev/ttyUSB1 1000000
```
This will start a `miniterm` session with two log files: one before the `default` filter (cleans some control chars etc...) and one after.
You will be prompted twice for log fil names, in the same order they were given on the command line.

##### Example 3
In a running session started as above, hit `ctrl+t ctrl+f` to edit current filters and enter for instance 'default': this will stop the log(s).
Do it again and enter `log default` to re-enable logging.

PR #270, opened 2017-09-16
fix an bug when run on windows XP.

leileigong:
as XP do not support API 'CancelIoEx'


kurtmckee:
@zsquareplusc @projectgus Please close this PR. No supported Python version runs on Windows XP.

PR #279, opened 2017-10-29
retry tcdrain on EINTR, fixes #278

### LINKED ISSUE #278: `termios.tcdrain` can fail with EINTR
We sometimes get an exception `termios.error: (4, 'Interrupted system call')` propagating out of pyserial's `serial.serialposix.flush`.

`tcdrain` can fail with EINTR if some signal was handled, see http://man7.org/linux/man-pages/man3/tcdrain.3p.html .

We are handling this application-side by catching `termios.error` and retrying. But since https://github.com/pyserial/pyserial/blob/master/serial/serialposix.py#L514 it looks like the idea is that pyserial should be handling EINTR, perhaps a retry for flush/tcdrain should be added?

Note: none of the other tc* functions appear to return EINTR.

Note2: Python 3.5 apparently retries a lot of syscalls https://www.python.org/dev/peps/pep-0475/ automatically, but from glancing at cpython's 3.5 branch it doesn't look like termios.tcdrain was given this treatment.

edmund-huber:
I'm still not sure retrying is the appropriate thing to do, but here you go.


kurtmckee:
@zsquareplusc @dpgeorge Please close this PR in favor of #830, which resolves merge conflicts and resolves Python 2-only syntax.

Issue #294, opened 2017-12-11
Support URLs with fully qualified package names


jamessynge:
In support of issue #292, it would help if `serial_for_url` would accept protocols that were composed of fully qualified package names. For example:

`serial.serial_for_url('path.to.test://')`

This would mean search in package `path.to` for module `protocol_test.py`. This would NOT require that the user add the package to be searched to `serial.protocol_handler_packages`, but would instead search only `path.to`.

zsquareplusc:
Not sure if that is a good idea. It would mean that a simple serial port name could cause code execution in other modules. Which could cause security issues. While most modules will guard code execution if the typical `if'__name__ == ' __main__':...`, not all do. I've found test scripts installed by modules that run directly.

jamessynge:
I'm surprised by that objection, given that we're talking about Python, which is inherently open source (i.e. you can read it, and can import or run anything you like). But, perhaps I'm not coming from the right perspective. So, how is this:
```python
serial.protocol_handler_packages.append('path.to')
serial.serial_for_url('test://')
```
Different from:
```python
serial.serial_for_url('path.to.test://')
```
The only thing I can imagine is that you're saying it is different for the case where the end user supplies a url parameter, but doesn't have ready access to the code. This mechanism would allow the user to command the import of an arbitrary module. Correct?

zsquareplusc:
The difference is when an application uses pySerial. In that case, the user does not care if it's Python or anything else. The comport name will be saved in some config file and no one would expect that this alone could import any Python module. If the source is available or not makes no difference, you just would not expect to be able to import arbitrary Python modules by the name of a serial port.

In the first case, url handler modules are explicitly registered for that purpose. So it can be expected that they are reasonably safe.

Issue #292, opened 2017-12-09
Support protocol handlers in miniterm


jamessynge:
I'd like to test my protocol handler using miniterm. For example:

`miniterm.py my_scheme:// 9600`

This fails because the code needs to add the appropriate package to `serial.protocol_handler_packages`. So, it would be great to add the required package via command line; for example:

`miniterm.py --pkg=my.handlers.package my_scheme:// 9600`

zsquareplusc:
wouldn't it work to set `PYTHONPATH`?

jamessynge:
It doesn't work properly, IIUC, if the handler is in a non-top-level package (i.e. the parent directory of the handler is also a package). In particular, if the handler needs to perform imports of other packages in its source tree (e.g. `import my.handlers.sibling` or `import ..sibling`), this would require that PYTHONPATH have two directories in it, the root of the source tree and the parent of the such as:

- `/root/of/source/tree`
- `/root/of/source/tree/my/handlers/package`

And, again IIUC, this means that sys.modules can easily end up with two imports of the same package but at different paths. Do you concur?

jamessynge:
I held off on this because I had offered PR #295, but since that is rejected (for what seems to be a good reason), I today tried to use a serial handler by adding its directory to PYTHONPATH, but this doesn't work because only pages in `serial.protocol_handler_packages` are searched:
```
protocol_handler_packages = [
    'serial.urlhandler',
]
...
        # if it is an URL, try to import the handler module from the list of possible packages
        if '://' in url_lowercase:
            protocol = url_lowercase.split('://', 1)[0]
            module_name = '.protocol_{}'.format(protocol)
            for package_name in protocol_handler_packages:
                try:
                    importlib.import_module(package_name)
                    handler_module = importlib.import_module(module_name, package_name)
```
So, would you consider a PR that adds a command line option for adding a package to  `serial.protocol_handler_packages`?

PR #295, opened 2017-12-11
Add support to serial_for_url for handlers with fully qualified package names.

jamessynge:
Addresses #294, and probably fully addresses #292.


zsquareplusc:
Not sure if that is a good idea. It would mean that a simple serial port name could cause code execution in other modules.

PR #303, opened 2018-01-12
Skip link detection in Windows

aivarannamaa:
If `p` is a Serial over Bluetooth link, then `os.path.islink(p)` may take several seconds.

In two of my Windows 10 computers following code used to take 5 seconds:

```
from serial.tools.list_ports import comports
for p in comports():
    print(p)
```
I found out that the delay was caused by `os.path.islink('COM5')`. COM5 is one of the "Bluetooth over serial" ports backed by LogiLink Bluetooth dongle.

After my change it takes 0.01 seconds.

If I'm not mistaken, then `list_ports_windows` can't list symlinks anyway, so this time consuming check can be turned off for Windows.

### DIFF
--- a/serial/tools/list_ports_common.py
+++ b/serial/tools/list_ports_common.py
@@ -11,7 +11,7 @@ import re
 import glob
 import os
 import os.path
-
+import platform
 
 def numsplit(text):
     """\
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Right, that could be skipped.

What I do not like is that this introduces a new method to check for the platform and makes that module platform dependent too. I guess it easier to add a parameter to the `ListPortInfo` constructor. I've implemented that.

PR #305, opened 2018-01-16
return b'' when connection closes on rfc2217 connection

Mispel:
Return an empty bytes object when the rfc2217 connection closes. This way the behavior is the same as closing a normal serial port.


zsquareplusc:
Yes, this makes sense, thank you.

PR #315, opened 2018-02-08
Use absolute import everywhere

kurtmckee:
This prevents Python 2.7 from attempting -- and failing -- to import system modules like sys and ctypes from the ``serial/`` directory.

At the company I work for, we're running Python 2.7 instances from a Windows share. Absolute imports in our production software have reduced system module import misses by 1000 per run. I'm going through third-party modules to further reduce the import misses. =)

Please let me know if you see anything out of place or need me to make any changes to get this patch accepted! I ran the unit tests on Windows and they continued to pass but let me know if I missed something. Thanks!

### DIFF
--- a/serial/__init__.py
+++ b/serial/__init__.py
@@ -7,6 +7,8 @@
 #
 # SPDX-License-Identifier:    BSD-3-Clause
 
+from __future__ import absolute_import
+
 import sys
 import importlib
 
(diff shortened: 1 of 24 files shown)


zsquareplusc:
Looks like i'm already using absolute imports anyway so the `from __future__ import absolute_import` does not do any harm... now that only Python 2.7 and newer is supported (it would probably not have worked for older Python releases).

Thank you.

Issue #124, opened 2016-06-04
DTR control line toggles unintentionally when opening port, dsrdtr state incorrect


gnbl:
Python 3.5.1 on Windows 8.1 64 bit, hardware is Silabs CP2102 USB-UART bridge with current drivers. pySerial upgraded from 3.0.1 to current 3.1 via pip3 (as admin)
```
>pip show pyserial
---
Metadata-Version: 2.0
Name: pyserial
Version: 3.1
Summary: Python Serial Port Extension
Home-page: https://github.com/pyserial/pyserial
```

Invoked with 
`import serial
ser = serial.Serial('COM3', 38400)`
causes the DTR line to change (at least once) so my connected hardware is reset. This does not happen if the port is opened with another software, HTerm (http://www.der-hammer.info/terminal/) which also allows individual control of these lines.

Testing

```
import serial
ser = serial.Serial()
ser.baudrate = 38400
ser.port = 'COM3'
print(ser)
sys.exit()
```

results in 

```
opening serial port...
Serial<id=0x40979e3860, open=False>(port='COM3', baudrate=38400, bytesize=8, parity='N', stopbits=1, timeout=None, xonxoff=False, rtscts=False, dsrdtr=False)
```
But it does toggle the control line(s) despite the "False".

See also http://stackoverflow.com/questions/15460865/disable-dtr-in-pyserial-from-code

But 
```
print("opening serial port...")
import serial
ser = serial.Serial()
ser.baudrate = 38400
ser.port = 'COM3'
print(ser)
ser.setDTR(0)
print(ser)
#sys.exit()
ser.open()
```

works.
The method is deprecated, however: http://pyserial.readthedocs.io/en/stable/pyserial_api.html#serial.Serial.setDTR 

So let's try again with the current API:

http://pyserial.readthedocs.io/en/stable/pyserial_api.html#serial.Serial.dtr

```
print("opening serial port...")
import serial
ser = serial.Serial()
ser.baudrate = 38400
ser.port = 'COM3'
print(ser)
ser.dtr = 0
print(ser)
ser.open()
```

and `ser.dtr = None` works as well!

So something must be wrong with the initial state.

May be related to https://github.com/pyserial/pyserial/issues/89

But dev seems to have shot down a similar issue just last year:
https://sourceforge.net/p/pyserial/bugs/176/#2ce7

Please fix or mention in the docs. Thanks!

gnbl:
I have changed the title because I have misunderstood the arguments:
[...]

PR #316, opened 2018-02-12
POSIX: Added support for setting and clearing the HUPCL flag.

sria91:
This tries to provide a means to fix the issue pointed out in #124.

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -260,6 +260,7 @@ class Serial(SerialBase, PlatformSpecific):
         if self.is_open:
             raise SerialException("Port is already open.")
         self.fd = None
+        self._hupcl_state = None
         # open
         try:
             self.fd = os.open(self.portstr, os.O_RDWR | os.O_NOCTTY | os.O_NONBLOCK)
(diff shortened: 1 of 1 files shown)


zsquareplusc:
OK, so this lets the user control the `HUPCL`state on posix but it still has to be called "manually" i.e. the users code is platform dependent. Maybe the state of this flag could be connected to the initial state of the rts/dtr properties: disable `HUPCL` if `rts` or `dtr` is assigned before `open()` is called? that way it could be similar on all platforms? => You have control over the control lines while opening the port, if they are assigned before `open()`.

JamesTheAwesomeDude:
There may be more to this than just `hupcl`, unfortunately.

For instance, running the following code:

	stty -a -F /dev/ttyACM0 | grep --only-matching -- '-hupcl' ; sleep 10; python3 -c 'from time import sleep;fd=open("/dev/ttyACM0");sleep(8)' & (sleep 3; tty -a -F /dev/ttyACM0 | grep --only-matching -- '-hupcl')

which:

* Checks the status of `hupcl` on `/dev/ttyACM0`
* (Waits 10s for the device to finish rebooting after a potential reset)
* Opens the device using `open` in Python,
*and, while it's open:*
  * Waits 3 seconds
  * Checks the status of `hupcl` again
  * Waits 5 more seconds, then
* Closes it in Python using `close`

When I ran it, reset the device during Python's `open` **despite `-hupcl` being output _twice_** (once for each check: the pre-check, and the check while Python was holding the file open.)

So, this patch will probably not fix #124 in any case. While it may be necessary, it seems it's certainly not sufficient to disable `HUPCL` to fix that bug.

PR #321, opened 2018-03-09
Added option for using comport device interface

82phil:
Provides an option to use the Device Interface Class GUID_DEVINTERFACE_COMPORT. This hopefully provides all comports including those not listed under the "Ports" Device Setup Class.

Verified operation using FTDI USB-UART converters and Com0Com Virtual Serial Ports.

### DIFF
--- a/serial/tools/list_ports_windows.py
+++ b/serial/tools/list_ports_windows.py
@@ -133,26 +133,24 @@ KEY_READ = 0x20019
 
 # GUID Class for Comport Device Interface - 86E0D1E0-8089-11D0-9CE4-08003E301F73
 GUID_DEVINTERFACE_COMPORT = GUID(
-        0x86e0d1e0L,
-        0x8089,
-        0x11d0,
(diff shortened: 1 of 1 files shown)


zsquareplusc:
This looks interesting, but who provides that flag? With this pull request the caller of `comports()` could o use the flag on Windows but when and how about cross platform code...

Would it be an option to scan both device classes and merge those lists to remove duplicates?

82phil:
@zsquareplusc I was not entirely sure if this was a desired behavior for all to list all comports, even those outside of the Ports Class (ex. Modem or Virtual Ports).

- Do you want to have the option added to common code, similar to include_links? Or no option and just merge the results from the two?
- I think I could merge the results from both based on the hardware_id. I was reviewing MSDN and did find an interesting note that "... a serial port is a COM port only if it has a COM port device interface associated with it." [External Naming of COM Ports](https://msdn.microsoft.com/en-us/library/windows/hardware/ff546502). However I am not sure if this is true for every serial device out there.

82phil:
@zsquareplusc Updated to run through both classes and remove duplicates

PR #324, opened 2018-03-16
Fix miniterm constructor defaults for exit_character and menu_character variables

carlosperate:
The MiniTerm.writer() method expects the exit_character and menu_character in character format instead of integer.
The main() function works because both member variables are overwritten with the correct format, any other programs using the default values from the constructor won't be able to exit or view the menu.


zsquareplusc:
Thank you.

PR #327, opened 2018-03-19
miniterm.py: use exclusive access for native serial ports by default

### LINKED ISSUE #326: miniterm.py does not use exclusive access for native serial ports
(The below applies to Linux / POSIX; haven't checked how it behaves on Windows)

Running `miniterm.py` while another process already has the same native serial port open will cause both processes to receive only garbage (or more accurately each will only receive parts of the data), even if the other process uses `flock()` for exclusive access. Since pyserial 3.3 the API offers the `exclusive` option to avoid several processes having the same native port open at the same time. However `miniterm.py` doesn't use this API, neither by default nor via an option.

silbe:
Fixes #326. miniterm.py now uses exclusive access for `serial.Serial` instances by default. The new option `--non-exclusive` can be used to override the default, disabling exclusive access for native ports. No change for non-native ports.

Tested on Linux (Python 3.4.2).

### DIFF
--- a/serial/tools/miniterm.py
+++ b/serial/tools/miniterm.py
@@ -809,6 +809,13 @@ def main(default_port=None, default_baudrate=9600, default_rts=None, default_dtr
         default=default_dtr)
 
     group.add_argument(
+        "--non-exclusive",
+        dest="exclusive",
+        action="store_false",
+        help="disable locking for native ports",
+        default=True)
+
+    group.add_argument(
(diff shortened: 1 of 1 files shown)


silbe:
@zsquareplusc Is there anything you would like me to change before this can go in? I've been using it on several systems (all Linux though) for over a month now without any issue.

silbe:
Great, thanks!

PR #338, opened 2018-05-04
Read bus reported name

sheppard74:
Added the bus reported name into info.name. This is returning the same name as linux does. It may need some testing on windows XP, since it isn't supported. I didn't put into description because of this and for compatibility.

### DIFF
--- a/serial/tools/list_ports_windows.py
+++ b/serial/tools/list_ports_windows.py
@@ -36,6 +36,7 @@ LPCTSTR = ctypes.c_wchar_p
 PCTSTR = ctypes.c_wchar_p
 PTSTR = ctypes.c_wchar_p
 LPDWORD = PDWORD = ctypes.POINTER(DWORD)
+PULONG = ctypes.POINTER(ULONG)
 #~ LPBYTE = PBYTE = ctypes.POINTER(BYTE)
 LPBYTE = PBYTE = ctypes.c_void_p        # XXX avoids error about types
 
@@ -48,7 +49,7 @@ class GUID(ctypes.Structure):
         ('Data1', DWORD),
         ('Data2', WORD),
         ('Data3', WORD),
-        ('Data4', BYTE * 8),
+        ('Data4', ctypes.c_ubyte * 8),
     ]
 
     def __str__(self):
@@ -59,7 +60,6 @@ class GUID(ctypes.Structure):
             ''.join(["{:02x}".format(d) for d in self.Data4[:2]]),
             ''.join(["{:02x}".format(d) for d in self.Data4[2:]]),
         )
-
 
 class SP_DEVINFO_DATA(ctypes.Structure):
     _fields_ = [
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Reading the description from the descriptor is indeed interesting. But i think it should set `.description` and not `.name`.

A drawback will also be that the names won't match with other port listings from windows (e.g. device manager or other applications). Maybe a switch will be needed or include two devices with both names in the list if include_links is set.

sheppard74:
I think, creating two devices with both names could be a nice solution. It won't break anything for driver using .inf for the name and it will give a way to find specific device into generic usb driver.

sheppard74:
Just remember that in posix it have been put in .description like this (usb.product - usb.cdcinterface). As shown here:

ABB CORESENSE M10 - ABB CORESENSE M10 SCADA
ABB CORESENSE M10 - ABB CORESENSE M10 SB

SJChannel:
I like this change, but I have a different opinion about where the information should be placed.  I believe that under Windows the "bus reported device description" should be stored into the "product" attribute of the ListPortInfo.  That would make it consistent with pyserial's behavior under Linux and MacOS.  See the attached file showing how the attributes are filled in on all three platforms currently.

In my opinion, the most important pieces of information for identifying a USB device are the vendor ID (VID), product ID (PID), manufacturer string, and product string.  Many USB devices use interface chips such as the Microchip MCP2221A USB-to-serial converter.  Generally, the VID and PID reported by these chips are immutable.  For example, the MCP2221A reports the VID for Microchip and a PID of 0x00dd.  The manufacturer and product **strings**, however, can be reprogrammed to any desired values.  This is very useful for distinguishing between all of the different products that use the same interface chip.  In the attached output, you'll see that pyserial under MacOS and Linux correctly reports the actual manufacturer and product strings returned by the device.  Under Windows it should do the same thing, to the extent that is possible.

Under Windows 10, the product string returned by the device is recorded as the "bus reported device description".  Pyserial should put that into the "product" attribute and (maybe) into the "description" attribute, as well.  That would make it consistent across platforms.  Unfortunately, I haven't found a way to retrieve the manufacturer string reported by the device under Windows.

[listports.txt](https://github.com/pyserial/pyserial/files/4339915/listports.txt)

sheppard74:
Following your "listports.txt" output, I agree that it has to be put in product. Tomorrow, I will take a look with my devices if I have the same behavior as you.

I will look for manufacturer if I can do something by the same time.

For description, it is more complicated, if program already use it to find their devices in Windows, it will break their code. Then, if we do, where should we put the "Microsoft name"?

SJChannel:
I agree, the description field is problematic.  Under Windows, "product" is currently set to None, so it shouldn't break anything to set it.  Maybe it's best to leave "description" as it is, at least for now.

sheppard74:
Ok, I did some test with my device, here is my findings.

In linux:
product=usb.productString
manufacturer=usb.manufString
interface=usb.interfaceString
desc=usb.productString - usb.interfaceString

In windows:
product=inf product string
manufacturer=none
interface=none
desc=inf description

Since I've built a device with 3 CDC, I need a mean to know which driver is which port. And, the CDC mechanism can return a different string for each CDC into interfaceString, not productString.

So, I need to move it from "name" to "interface". For product and manufacturer, usbview show how to extract this from usb descriptor. It sounds a lot of work to do it by the way.

sheppard74:
This web site show how to do it: https://stackoverflow.com/questions/24756591/python-to-get-usb-descriptor

SJChannel:
@sheppard74 Thanks for finding that stackoverflow solution.  After making a few python2->3 changes, it works for me (miniconda with python 3.76 and pywin32 227).  I'm attaching my modified version of the stackoverflow code and the output from it.  Strangely, the third item listed (Focaltech Systems) sometimes reports "ERROR: no String Descriptor ...".

It does seem like a lot of work to do all of this.  I personally would be satisfied with your original solution, storing just the product string into the "product" attribute.
[get_usb_descriptor.py.txt](https://github.com/pyserial/pyserial/files/4345346/get_usb_descriptor.py.txt)
[output.txt](https://github.com/pyserial/pyserial/files/4345347/output.txt)

LunaHub:
Hi, 
I am trying to recognize the port connected to an Arduino. It would therefor be really great if I could get the _Bus reported device desciption_ as seen under the windows device manager (which shows "Arduino Nano"). Any fix for this? I am not sure how to use the code above :( Non of the .description, .product or .manufacture returns any usefull information.

Issue #319, opened 2018-03-01
namespace conflict with https://pypi.python.org/pypi/serial/0.0.41?


dnahmanpix:
Both pyserial and serial-0.0.41 (which is a dependency of requests, and other packages) both install to the same 'serial' directory in site-packages and clobber each other.  Is there a sophisticated way to make pyserial install to a pyserial namespace without forking and changing the code itself?

osterwood:
I am having the same issue.  Have you found any work arounds?

dnahmanpix:
Alas, no.  I ended up forking the project and maintaining it on our internal pypi server.

gowhari:
Owner of `serial` package kindly renamed it to `sob` to stop this conflict:
https://bitbucket.org/davebelais/serial/issues/2/namespace-conflict

It still needs some changes on pypi to fix the problem.

PR #340, opened 2018-05-07
setup.py: Add install-time check for conflicting 'serial' package

projectgus:
Addresses https://github.com/pyserial/pyserial/issues/319

This is derived closely from the equivalent check applied by @davebelais to
'serial', here:
https://bitbucket.org/davebelais/serial/commits/363abb1c5d92dfdb3f1b3e3af70fee28bd05e053#Lsetup.pyT1

This doesn't fix issues in code which uses 'pyserial', in cases where 'serial'
was accidentally installed on the system instead. But it prevents issues caused
when both packages are installed at once.

Thanks for all your hard work on pyserial. It's a great package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,24 @@ try:
 except ImportError:
     from distutils.core import setup
 
+try:
+    import pkg_resources
+    try:
+        pkg_resources.get_distribution('serial')
+        raise RuntimeError(
+            """This package cannot be installed alongside `serial`, due to a module name conflict.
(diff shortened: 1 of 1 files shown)


dpgeorge:
@projectgus this can probably be closed, considering #811.

kurtmckee:
@zsquareplusc @projectgus Please close this PR.

projectgus:
Huh, I forgot I ever submitted this. 

Will close. Might revisit at some point, as this was a constant bugbear for esptool users back in the day.

PR #354, opened 2018-06-13
Make ListPortInfo hashable

vlovich:
Makes it behave as consistently within unordered data structures like sets/hashes.


zsquareplusc:
Makes sense, thank you.

PR #356, opened 2018-06-17
Documented read_until function

jdpatt:
I end up writing my own flavor of read_until every time to line up with telnetlib.  Didn't know it was apart of pyserial already!  Documented the function.  I also tweaked the parameter name to match telnetlib but it is not affecting anything if that is left as terminator.

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -649,19 +649,19 @@ class SerialBase(io.RawIOBase):
         """
         return self.read(self.in_waiting)
 
-    def read_until(self, terminator=LF, size=None):
+    def read_until(self, expected=LF, size=None):
         """\
-        Read until a termination sequence is found ('\n' by default), the size
+        Read until an expected sequence is found ('\n' by default), the size
         is exceeded or until timeout occurs.
         """
-        lenterm = len(terminator)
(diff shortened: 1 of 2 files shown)


serial/serialutil.py
```diff
@@ -649,19 +649,19 @@ def read_all(self):
         """
         return self.read(self.in_waiting)
 
-    def read_until(self, terminator=LF, size=None):
+    def read_until(self, expected=LF, size=None):
```
zsquareplusc: renaming the parameter may break some code, but it makes sense to align the name with the function from telnetlib

jdpatt:
Looks like I should have escaped the '\\';  Just noticed on read the docs that it rendered as (‘n’ by default) instead of ('\\n’ by default)  Thanks for the awesome package!

serial/serialutil.py
bgottula: Confirming that this broke some code. Does this project not use semantic versioning? This rolled out with version 3.5b0 if I follow the history correctly, yet it included incompatible API changes such as this. The changelog lumps this in with "doc updates" but changing a parameter name is not only for documentation.

PR #358, opened 2018-06-20
Properly escape backslash.

jdpatt:
Missed this on the last pull request.  Currently rendering as (‘n’ by default) instead of ('\n’ by default)


zsquareplusc:
:smile:

PR #360, opened 2018-06-24
Add list_ports and miniterm entry points

ksunden:
Note, this also removes support for bare `distutils` in setup.py

`setuptools` is now so ubiquitous (pip uses it even if it is not imported directly) that I do not see this as a problem, but I can refactor to include if it is desired to keep this functionality.

entry_points are a tool specific to `setuptools`

I also had to guess at the next version for the ``versionchanged`` in the docs

### DIFF
--- a/documentation/shortintro.rst
+++ b/documentation/shortintro.rst
@@ -93,7 +93,7 @@ Testing ports
 =============
 Listing ports
 -------------
-``python -m serial.tools.list_ports`` will print a list of available ports. It
+``list_ports`` will print a list of available ports. It
 is also possible to add a regexp as first argument and the list will only
 include entries that matched.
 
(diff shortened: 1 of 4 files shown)


ksunden:
Just realized that there is a previously opened PR which includes this (#64) (though that does not update the docs as well)

To answer the question of "why add list_ports?", I used this just the other day and had to remember the `python -m` command. While I remembered it after just a brief moment, I know others in my research lab who do not have that memorized (nor use `python -m` at all) but who may wish to know which ports are available on a given machine.

I could write an alias/script myself, but I wanted to submit the utility for others to use more broadly.

`entry_points` provide a standard way of providing tools which are available from the command line easily and in a cross platform way, where users do not need to remember a long command.

zsquareplusc:
I am not a big fan of spamming the PATH with lot of little tools. e.g. on a Linux, pyserial may be pre-installed but the user would not expect to get lots of command line tools-

Maybe if a prefix is considered, e.g. `pyserial-miniterm` and `pyserial-list_ports`...

PR #364, opened 2018-07-13
macOS: rework list_ports to support unicode product descriptors.

torque:
A number of Arm-based microcontrollers, such as the kinetis-based Teensy 3 family, provide a native USB interface that allows the user to control a number of attributes, such as the product descriptor string. However, macOS exhibits some weird behavior where it mangles specifically the product descriptor string (and as far as I can tell, only the product descriptor string) on USB serial devices (this also does not seem to happen with other types of USB device). The mangling, which replaces all non-alphanumeric characters with underscores, appears to happen somewhere deep in the kernel's BSD serial subsystem. In order to work around this, the device's product name can be queried with a different IOKit call, specifically `IORegistryEntryGetName`, which appears to be aliased directly to the (unmangled) USB product descriptor string for all USB devices.

Note: I'm not sure what versions of macOS pyserial aims to support, but I have tested this (briefly) on both 10.11 and 10.13.

This PR also attempts to make a few correctness fixes to the way pyserial interfaces with IOKit. These have been squashed into the same commit for the sake of brevity.

### DIFF
--- a/serial/tools/list_ports_osx.py
+++ b/serial/tools/list_ports_osx.py
@@ -35,25 +35,40 @@ kIOMasterPortDefault = ctypes.c_void_p.in_dll(iokit, "kIOMasterPortDefault")
 kCFAllocatorDefault = ctypes.c_void_p.in_dll(cf, "kCFAllocatorDefault")
 
 kCFStringEncodingMacRoman = 0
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thank you

PR #368, opened 2018-07-18
Add support for green threads

TaQk:
Today I noticed that the serial.read() causes a deadlock when the threading library is monkey_patched by e.g. eventlet (threads in cooperative mode against parallel) even when timeout is specified. Example:
```
STOP = False

def test():
    global STOP
    sh = serial.Serial('COM1', timeout=.1)
    while not STOP:
        sh.read(1)

threading.Thread(target=test, daemon=True).start()
time.sleep(1)
STOP = True
#thread will be killed

#but in case of:
eventlet.monkey_patch()
threading.Thread(target=test, daemon=True).start()
#program will never reach the point below!
STOP = True
```

The solution is a little bit tricky. time.sleep(0) won't cause any time delay but allows to switch context and solves the problem.

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -36,6 +36,7 @@ import select
 import struct
 import sys
 import termios
+import time 
 
 import serial
 from serial.serialutil import SerialBase, SerialException, to_bytes, \
@@ -538,6 +539,10 @@ class Serial(SerialBase, PlatformSpecific):
                     raise SerialException('read failed: {}'.format(e))
             if timeout.expired():
                 break
+
+        # Give an opportunity to switch a context. It prevents deadlock if 
+        # a greenthread is used (e.g. after eventlet.monkey_patch()).  
+        time.sleep(0)
         return bytes(read)
 
     def cancel_read(self):
--- a/serial/serialwin32.py
+++ b/serial/serialwin32.py
@@ -298,6 +298,10 @@ class Serial(SerialBase):
                 read = bytes()
         else:
             read = bytes()
+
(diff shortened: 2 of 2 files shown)


ffleandro:
This did not work for me in a Raspberry Pi.

TaQk:
Could you explain? What exactly didn't work?

ffleandro:
I've debugged and the serial communication does not read any data inside a green thread, I guess the thread never gets notified that new data is available to process.

The same code outside a thread works, but inside a green thread, for example, with `eventlet.monkey_patch` it does not receive any packet.

If using a normal `threading.thread` without eventlet, everything works fine.

zsquareplusc:
I don't see how a (green)thread change at that point would help, the `sleep(0)` calls are right in front of the return. But those calls could have a negative impact (wasting CPU time) for normal usage :/

note that there is also `pySerial-asyncio` which supports lightweight threading.

for the example above, i'd expect that it works, without any patches in pySerial, if you add the `eventlet.sleep(0)` in the `test` function's while-loop.

You could also use `timeout=0` for truly non-blocking reads on the serial port (though the CPU load may rise because of more empty reads. to avoid this either `sleep` longer when zero bytes have been read or use `select` - at least under posix it is possible to wait for serial ports the same way as to wait for sockets)

kurtmckee:
@zsquareplusc @dpgeorge Please close this PR without merging.

eventlet heavily discourages using their library in favor of asyncio. [[source](https://eventlet.readthedocs.io/en/latest/)]

dpgeorge:
Yes, `asyncio` should be used instead.

Issue #88, opened 2016-03-09
inter_byte_timeout makes no difference


velle:
I suspect I am using it the wrong way, so I'm not sure whether it is proper that I write here. However, I could not make the inter_byte_timeout work, and I could not find help in other places. 

I have a very simple Python script, that has inter_byte_timeout set. It makes no difference though (the way I am using it). 

```
import serial
ser = serial.Serial('/dev/ttyACM0', inter_byte_timeout=1.0, timeout=3.0)
s = ser.read(50)
print(s)
```

I have a very simple Arduino script that writes to serial. 

```
void setup() {
  Serial.begin(9600);
  while(!Serial);
  for (int i=0;true;i++){
    Serial.printf("=== iteration %d ===\n", i);
    delay(1000);
  }
}

void loop() {
}
```
The result is that the Python script writes every 3 seconds, not every 1 second. Have I misunderstood this?

It's version 3.0.1.

zsquareplusc:
The timeout and the value used to test are a bit close and both systems may not agree how long a second is, exactly. It would make sense to use a slightly larger delay on the Arduino (or a lower one in pySerial). e.g. `inter_byte_timeout=0.5` and `delay(1000)`

The inter byte timeout is done by the driver or even the USB-Serial adapters firmware. It may also be that some devices to not support this properly.

rbertoche:
I don't get zsquareplusc comment, velle describes clearly that the feature isn't working.
I get the same results, pyserial seems to ignore completely that timeout
Could you tell me how are you testing it to make it work?
I have tons of scripts that depend on the older pyserial versions where timeout was actually a bad name for the behavior inter_byte_timeout should gives us now. But then it worked and now I guess it doesn't...
Thank you

zsquareplusc:
`interCharTimeout` was renamed to `inter_byte_timeout`, still doing the same, as it was in older versions. `timeout` has not changed at all.

The `timeout` is implemented with OS functions, depending on the platform, e.g. with `select`'s  timeout parameter and not with settings of the driver.

In contrast to `inter_byte_timeout` which is a setting of the driver and if that is working properly also depends on the driver and the capabilities of the chip (e.g. USB serial converter, serial port emulations). It would not surprise me if you find devices not doing that one properly.

rbertoche:
Got it,
[...]

PR #373, opened 2018-08-06
fix inter_byte_timeout

mahatma1:
see issue #88

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -538,7 +538,7 @@ class Serial(SerialBase, PlatformSpecific):
                     raise SerialException('read failed: {}'.format(e))
             if timeout.expired():
                 break
-            else if self._inter_byte_timeout is not None and self._inter_byte_timeout > 0:
+            elif self._inter_byte_timeout is not None and self._inter_byte_timeout > 0:
                 #atleast one char received --> overwrite timeout by setting it to inter_byte_timeout
                 timeout = Timeout(self._inter_byte_timeout)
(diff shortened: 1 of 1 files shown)


serial/serialposix.py
```diff
@@ -538,6 +538,10 @@ def read(self, size=1):
                     raise SerialException('read failed: {}'.format(e))
             if timeout.expired():
                 break
+            elif self._inter_byte_timeout is not None and self._inter_byte_timeout > 0:
+                #atleast one char received --> overwrite timeout by setting it to inter_byte_timeout
+                timeout = Timeout(self._inter_byte_timeout)
```
zsquareplusc: are you sure this is needed? the `inter_byte_timeout` is actually set on the port using `vtime`, so it should already have elapsed that time and a simple `break` -- `if self._inter_byte_timeout is not None` -- would be enough here.

mahatma1: it's been a while since I played around with it but as far I can remember I wasn't able to get the inter_char/inter_byte timeout working on a raspberry pi without my fix - I am pretty sure that I  tried out different values for the timeout (milliseconds, seconds etc as mentioned in #396).

xrombik: > enough her

You **inter_byte_timeout** is useless. It does not worked on any platforms.

chrisr710:
The fix above worked for me

RossSmyth:
Yeah this should be merged because it works.

PR #374, opened 2018-08-07
Removed Errant `open()` Call in Context Mgr Docs

WillAyd:
If you try to execute this code as is you will get:

```python
SerialException: Port is already open.
```

since the context manager already opens the port on entry

### DIFF
--- a/documentation/shortintro.rst
+++ b/documentation/shortintro.rst
@@ -49,7 +49,6 @@ Also supported with :ref:`context manager <context-manager>`::
     with serial.Serial() as ser:
         ser.baudrate = 19200
         ser.port = 'COM1'
-        ser.open()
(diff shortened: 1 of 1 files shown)


keelung-yang:
"Already open" is a bug since port name not supplied while entering context mgr. See my pull request #371

WillAyd:
Hmm interesting. In my case I was specifying the port name as part of the context manager construction and was expecting it to be open on entry (which it was) but I can see if you don't specify a port number that you would need to explicitly call that again.

Perhaps it would make the most sense then to just make the port number and baudrate required arguments for the context manager (plus anything else that maybe needs to be there)? It seems like the usage pattern will be somewhat confusing if there are multiple ways to go about the actual opening

PR #386, opened 2018-10-03
rfc2217/close(): fix race condition

jfleschExtFlowbird:
When stopping the read loop, do not set `self._thread = None` inside the thread itself as there may be a race-condition with the method close().

The method close() closes the socket, which stops the read loop. When the read loop stops, it set `self._thread` to `None`, but if it sets it to `None` while close() is right between between the execution of `if self._thread:` and `self._thread.join()`, close() will raise an AttributeError.


zsquareplusc:
Thank You

PR #389, opened 2018-10-24
Replaced RFC 2271 with 2217 in docs & comments

lowell80:
Fixed transposed RFC 2217 (was 2271)  

Updated docs and comments incorrectly referencing 2271 (SNMP).  No code change.


zsquareplusc:
Thank you

PR #400, opened 2018-12-21
Add bytesize and stopbits argument parser to tcp_serial_redirect

smeng9:
Enhancement. Allow users to set these two options.


zsquareplusc:
looks OK, Thank you!

PR #409, opened 2019-01-28
miniterm.py:  Adds execute permissions.

fhgwright:
This has a suitable shebang line to be run directly as a program, but
needs execute permissions to make that actually work.

The same change could be made to list_ports*, but it's less important
there where command-line invocation is only for test purposes.

TESTED:
./miniterm.py now launches on macOS.  Other POSIX systems should
behave the same.  No idea what effect this has on Windows.


zsquareplusc:
I don't think that there is a reason to do this. My recommendation is to start these tools using `python -m serial.tools.miniterm`  and `python -m serial.tools.list_ports`.

And in case the script is installed via `setup.py` it takes care of making it executable.

(re windows: the execute flag is ignored as not supported by file system)

fhgwright:
Why is requiring extra typing considered a benefit?

Why is there a suitable shebang line if it's not intended to be run directly as a program?

Why is it OK to have it made executable via setup.py but not otherwise?

What possible downside is there to making it executable in the repo, as is normal practice in other projects with Python programs?

noahp:
It seems reasonable to make this change- note also that the `miniterm.py` script is installed to the `sys.prefix/bin` directory when installing pyserial as a package (i.e., `pip install pyserial` etc), and the executable flag is properly set there:

https://github.com/pyserial/pyserial/blob/master/setup.py#L99

fhgwright:
Ping.

kurtmckee:
@zsquareplusc @dpgeorge Please close this PR without merging.

Installing pyserial as a user (e.g. `pipx install pyserial`) or as a developer who has cloned the repo (e.g. `python -m pip install -e.`) both provide the `pyserial-miniterm` executable, which work reliably.

PR #411, opened 2019-01-29
Add a backend for Silicon Labs CP2110/4 HID-to-UART bridge.

Flameeyes:
These two chips implement UART access via a HID protocol, which can be
implemented purely in user space. The protocol is documented by Silicon
Labs AN434:
https://www.silabs.com/documents/public/application-notes/AN434-CP2110-4-Interface-Specification.pdf

The backend is implemented based on cython-hidapi
(https://github.com/trezor/cython-hidapi), making it OS-independent, if a
bit awkward.

### DIFF
--- a/documentation/url_handlers.rst
+++ b/documentation/url_handlers.rst
@@ -16,6 +16,7 @@ The function :func:`serial_for_url` accepts the following types of URLs:
 - ``hwgrep://<regexp>[&skip_busy][&n=N]``
 - ``spy://port[?option[=value][&option[=value]]]``
 - ``alt://port?class=<classname>``
+- ``cp2110://<bus>:<dev>:<if>``
 
 .. versionchanged:: 3.0 Options are specified with ``?`` and ``&`` instead of ``/``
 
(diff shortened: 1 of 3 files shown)


zsquareplusc:
could you move the complete code into `serial/urlhandler/protocol_cp2110.py`? I don't think that anyone would need to import `serial.cp2110`.

And there should probably be a note in `documentation/url_handlers.rst` mentioning the handler and that the `hidapi` module needs to be installed separately (i would not like to install that by default, but as option would be possible, via `extras_require={'cp2210': [hidapi']}` in `setup.py`).

I've also used HID here and did my own low-level access:
https://github.com/zsquareplusc/python-msp430-tools/blob/master/msp430/bsl5/hid.py#L84
hoever that does not (yet) support MacOSX

Flameeyes:
Thanks for the review!

Code moved and documentation and `extras_require` added.

I have implemented raw HID support myself before, and honestly it's sometimes nicer particularly with the ability to have a more rational path compared to hidapi. But the problem with that is that implementing select/timeout on it is more fragile than HIDAPI, which is why I'm not terribly keen on doing that here.

zsquareplusc:
Thank you

Issue #283, opened 2017-11-17
[windows] Serial Number Issues 


DuaneEllis-TI:
For USB composite devices, Pyserial cannot extract extract the serial number. 
What Pyserial needs to do is look at the *parent* device 

When Pyserial extracts the "szHardwareID" - it gets a strings like (see below)
What I am seeing appears to be a "device instance string"
References:

a)  https://docs.microsoft.com/en-us/windows-hardware/drivers/install/device-instance-ids
b) https://msdn.microsoft.com/en-us/library/windows/hardware/jj649944(v=vs.85).aspx

What is needed is this:

Step 1 - if you cannot find the serial number ...

Call the function:  CM_Get_Parent() - pass:  devinfo.DevInst  for the current device.

This gives you the PARENT device ID (this is an ephemeral number - meaning it changes when you reboot the machine, or plug & unplug devices)  

Step 2:

Call the function CM_Get_Device_ID() - pass the PARENT ID  (from above) and a pointer to a buffer to hold the USB string.

Step 3: Inspect the parent USB Vendor ID & USB Product ID - compare with your base child.

Case 1:  
   They do not match - you are done. You have the USB Hub the device is connected to.
   If a serial number exists, it is in the child... 
   Some do not have serial numbers, see COM 114

Case 2:
   They match - but you can't find the serial number - See COM134 as the example.
   Try the parent again... you should only have to go up 2 or 3 times - 4 at most.
   The serial number - if it exists is on the top most element of the matching device.

Case 3:
    You found the serial number - Great success, and Profit!

Below are several examples, note: in this example I am only traversing up 5 levels of parents.

Some notes:

1) it would be helpful if PySerial could identify the MI_NUMBER as an additional attribute
  This corresponds to the USB interface index, not the "NTH-in-some-order"
  It should be the USB interface index number this never changes.
  Rational:  Some devices have multiple serial ports and you need the NTH port...

2) Composite example : COM98 & COM99  are from the same composite device - two interfaces
    (It happens to be a TexasInstruments Launch Pad - they have 6 USB interfaces)

3)  COM96 is a standalone USB cable from FTDI

4) COM147 - is a device from Ember (now SiLabs) it is a non-composite device.
   This is plugged into the side of my laptop which as an (VID_8087 - aka: Intel) USB Hub

5) COM114 - has no serial number, it has a CP110x usb converter
   And it does not have a "MI_XX" number... it is a non-composite device.

[...]

PR #414, opened 2019-02-12
[Windows] Serial number support for composite USB devices

ryan-summers:
This pull request fixes https://github.com/pyserial/pyserial/issues/283 by adding in a device traversal up the USB device tree if a serial number is not detected at the original device node.

This allows composite USB devices to report correct serial numbers on windows.

### DIFF
--- a/serial/tools/list_ports_windows.py
+++ b/serial/tools/list_ports_windows.py
@@ -165,7 +165,7 @@ def get_parent_serial_number(child_devinst, child_vid, child_pid, depth=0):
 
     # Get the parent device instance.
     devinst = DWORD()
-    ret  = CM_Get_Parent(ctypes.byref(devinst), child_devinst, 0)
+    ret = CM_Get_Parent(ctypes.byref(devinst), child_devinst, 0)
 
     if ret:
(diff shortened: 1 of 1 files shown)


serial/tools/list_ports_windows.py
```diff
@@ -21,6 +21,7 @@
 from ctypes.wintypes import ULONG
 from ctypes.wintypes import HKEY
 from ctypes.wintypes import BYTE
+import winerror
```
zsquareplusc: does not seem to be in the standard library

ryan-summers: Thanks for pointing this out - you were correct. This was a module provided by `pywin32` that I (incorrectly) thought was part of the standard library on windows. I removed the dependency and added the error as a file-level constant similar to other windows errors handled here.

ryan-summers:
@zsquareplusc Is there anything else you would like to see with this PR?

hmaarrfk:
Thanks for working on this. I think we just hit this issue. I'll be versionning this code and testing it next week. Excited to see this pulled in!

zsquareplusc:
Thank you

PR #430, opened 2019-05-13
Add WaitCommEvent function to win32

ARF1:
This is helpful when a user wants to wait for some kind of communication event. E.g. listening for parity errors, etc.


zsquareplusc:
Simple enough, Thanks

PR #433, opened 2019-05-19
tools/list_ports_windows: Scan both 'Ports' and 'Modem' device classes

keith-packard:
Devices using the Window USB Modem driver use the Modem GUID instead
of the Ports GUID. Add the Modem class to the set of GUIDs to scan.

Signed-off-by: Keith Packard <keithp@keithp.com>


zsquareplusc:
Thank you

PR #441, opened 2019-06-20
reset_input_buffer() can hang on sockets

NJDFan:
If the remote end closes the socket, the socket object can report read-ready on the select call but the recv() always returns 0 bytes .  This empty recv leaves the select reporting ready, so the loop runs endlessly.


zsquareplusc:
Good catch. Thank you.

PR #468, opened 2019-12-03
Update serialutil.py

pierriko:
https://stackoverflow.com/questions/33734170/why-doesnt-python-auto-escape-in-doc
https://www.python.org/dev/peps/pep-0257/#specification

  For consistency, always use """triple double quotes""" around docstrings. Use r"""raw triple double quotes""" if you use any backslashes in your docstrings. For Unicode docstrings, use u"""Unicode triple-quoted strings""".

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -650,7 +650,7 @@ class SerialBase(io.RawIOBase):
         return self.read(self.in_waiting)
 
     def read_until(self, expected=LF, size=None):
-        """\
+        r"""\
         Read until an expected sequence is found ('\n' by default), the size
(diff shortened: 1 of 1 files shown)


zsquareplusc:
fixed by not using a backslash ;-)

the `\` at the beginning of the docstring is not required and most do not do this, as most tools processing docstrings will fix the uneven indentation if test starts on first line respectively remove an empty line if it does not.

the patch was probably intended to fix the `\n` in the text itself. but would have made the 1st `\` an error.

NickCrews:
Ah, I only say the first `\`, and not the `\n`. That fix looks good.

PR #475, opened 2019-12-28
Add SerialBase.read_between

jacobian91:
This adds functionality to read between two character sequences. Useful for making sure that a whole response is found.


jacobian91:
Removed Python 2.2/2.3 similar to PR #500 to pass Travis CI Checks.

PR #480, opened 2020-01-28
Handle FramedPacket.START bytes within a packet

jonathongrigg:
If we are inside a packet, the current logic skips over any bytes that match the START byte. I believe the correct behaviour would be to include all bytes since the first START byte up until the STOP byte, as proposed here

### DIFF
--- a/serial/threaded/__init__.py
+++ b/serial/threaded/__init__.py
@@ -100,7 +100,7 @@ class FramedPacket(Protocol):
     def data_received(self, data):
         """Find data enclosed in START/STOP, call handle_packet"""
         for byte in serial.iterbytes(data):
-            if byte == self.START:
+            if byte == self.START and not self.in_packet:
                 self.in_packet = True
             elif byte == self.STOP:
                 self.in_packet = False


NickCrews (changes requested):
Hi, thanks for this, this definitely seems like it's going in the right direction! I agree with you that `START` bytes within the packet should also be included.

However, this opens up another, related can of worms, and that is "what to do with STOP bytes received outside of a packet?" As of now we just call `handle_packet()` with an empty byte array, which seems OK, but maybe not great. I would think instead we should be calling `handle_out_of_packet_data()`. Therefore I propose this alternate structure, which also fixes this edge case, and I think is easier to understand than this 4-case if-else:

```
def data_received(self, data):
    """Find data enclosed in START/STOP, call handle_packet"""
    for byte in serial.iterbytes(data):
        if not self.in_packet:
            if byte == self.START:
                self.in_packet = True
            else:
                self.handle_out_of_packet_data(byte)
        else:  # We are in a packet
            if byte == self.STOP:
                self.in_packet = False
                self.handle_packet(bytes(self.packet)) # make read-only copy
                del self.packet[:]
            else:
                self.packet.extend(byte)
```
Any thoughts on this?

zsquareplusc:
Seeing an other start character is in my opinion an error (should not be contained in data). Therefore this would be an transmission error of some sort.

- We could assume that a previous packet was cut off and the new start is a true start of a new packet. In that case any in-packet data received up to here should be forgotten (or maybe handled by the `handle_out_of_packet_data`)
- A no-so-clean protocol might define that start characters may well be part of the data. In that case the byte should just be stored.

The current pySerial implementation does not handle either of these cases nicely.

When changing something, i'd probably go for the 1st variant, assuming a start is always a start of a new packet and discard previously received data, because this is a real thing when senders are stopped (mid packet) and restarted. Receiving the tail end of a packet never seen a start is a similar problem that can occur when cables are plugged in and out while communication is running.

PR #488, opened 2020-03-08
Add a log option

Mrbenoit624:
--log <filename>      store output in log files

### DIFF
--- a/serial/tools/miniterm.py
+++ b/serial/tools/miniterm.py
@@ -344,6 +344,7 @@ class Miniterm(object):
         self.serial = serial_instance
         self.echo = echo
         self.raw = False
+        self.log = ""
         self.input_encoding = 'UTF-8'
         self.output_encoding = 'UTF-8'
         self.eol = eol
@@ -375,6 +376,8 @@ class Miniterm(object):
     def start(self):
         """start worker threads"""
         self.alive = True
+        if self.log:
+            self.LogFile = open(self.log, 'ab')
         self._start_reader()
         # enter console->serial loop
         self.transmitter_thread = threading.Thread(target=self.writer, name='tx')
@@ -396,6 +399,8 @@ class Miniterm(object):
 
     def close(self):
         self.serial.close()
+        if self.log:
+            self.LogFile.close()
 
     def update_transformations(self):
         """take list of transformation classes and instantiate them for rx and tx"""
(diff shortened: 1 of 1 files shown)


noxobscura:
I would recommend the use of the logging module instead as it has a wide range of options and levels of granularity.

NickCrews:
> I would recommend the use of the logging module instead as it has a wide range of options and levels of granularity.

I agree, as `serial` is a library, if it does any logging, then it should be via the logging library. Then, clients who use the library should set log config, handlers, outputs, files, etc with the same logging library interface. It's pretty standard at this point, and much more flexible.

NickCrews:
> I agree, as `serial` is a library, if it does any logging, then it should be via the logging library. Then, clients who use the library should set log config, handlers, outputs, files, etc with the same logging library interface. It's pretty standard at this point, and much more flexible.

Well, I misunderstood a bit, actually miniterm is acting as its own standalone application here so `logging` isn't as needed, but it still is probably a good idea.

@zsquareplusc, would you be open to me adding `logging` library support to `serial` in general? I was thinking of both converting all of these `sys.stderr.write()` calls in the `miniterm` and `list_ports` apps to use logging, as well as actually going through the library code and adding logging calls that don't exist at this point. I have a feeling (with no evidence) that people might enjoy being able to turn on debug-level logging in serial for when they are debugging their applications.

zsquareplusc:
`miniterm` intentionally outputs extra info / interacts on `stderr`, so that its output could be redirected, e.g. with `miniterm /dev/ttyACM1 | tee output.txt`

to debug even lower level, using the `spy` URL handler is handy:

`miniterm spy:///dev/ttyACM1?file=output.txt`

so i do not immediately see a need for more logging.. except for a variant in the `spy` protocol to use `logging` instead of file/stdio.

NickCrews:
One problem with the `spy` protocol is that it can't be combined with the other protocols. Either you use spy to get debugging and logging, or you choose a different protocol, but for instance you can't currently log/debug a `socket://` protocol. In an ideal world I might think that logging should be an orthogonal choice to the protocol, so that logging could be used with whatever protocol you want.

I think it would be possible to maintain the outward interface of how the logging output happens, eg so that current uses of `miniterm /dev/ttyACM1 | tee output.txt` still work, but we still might mess up someone's work that depends on the internals of pyserial.

PR #494, opened 2020-03-30
Don't catch the SerialException we just raised

kalvdans:
In Python3 we get double tracebacks for this error:
```
  File "/home/chn/repo/pyserial/serial/serialposix.py", line 557, in read
    raise SerialException(
serial.serialutil.SerialException: device reports readiness to read but returned no data (device disconnected or multiple access on port?)

During handling of the above exception, another exception occurred:

  File "test/test_exceptions.py", line 26, in <module>
    test_unexpected_eof()
  File "/usr/lib/python3.8/unittest/mock.py", line 1348, in patched
    return func(*newargs, **newkeywargs)
  File "test/test_exceptions.py", line 21, in test_unexpected_eof
    s.read()
  File "/home/chn/repo/pyserial/serial/serialposix.py", line 566, in read
serial.serialutil.SerialException: read failed: device reports readiness to read but returned no data (device disconnected or multiple access on port?)
```
The patch moves the checking for EOF to after the IO block, resulting
in a much nicer traceback. This is the test script (Python3-specific):
```
from unittest import mock
import serial
import select
import os

@mock.patch('select.select')
@mock.patch('os.read')
def test_unexpected_eof(mock_read, mock_select):
    s = serial.Serial()
    s.is_open = True
    s.fd = 99
    s.pipe_abort_read_r = 98
    mock_select.return_value = ([99],[],[])
    mock_read.return_value = b''
    try:
        s.read()
    except serial.SerialException as e:
        if e.__context__: raise

if __name__ == '__main__':
    test_unexpected_eof()
```

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -548,16 +548,6 @@ class Serial(SerialBase, PlatformSpecific):
                 if not ready:
                     break   # timeout
                 buf = os.read(self.fd, size - len(read))
-                # read should always return some data as select reported it was
-                # ready to read when we get to this point.
-                if not buf:
-                    # Disconnected devices, at least on Linux, show the
-                    # behavior that they are always ready to read immediately
-                    # but reading returns nothing.
-                    raise SerialException(
-                        'device reports readiness to read but returned no data '
-                        '(device disconnected or multiple access on port?)')
-                read.extend(buf)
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Makes sense. Thank you

PR #495, opened 2020-03-30
Correct "interface" property on Linux hosts

johannesacco:
The "interface" property should be read from the "usb_interface_path"
instead of the "device_path" as the later one is a directory (e.g.
/sys/devices/pci0000:00/0000:00:14.0/usb1/1-2/1-2.3/1-2.3:1.0/ttyUSB1).

This has been tested together with an CP2105EK where it is important
to separate the serial ports from each other.

### DIFF
--- a/serial/tools/list_ports_linux.py
+++ b/serial/tools/list_ports_linux.py
@@ -59,7 +59,7 @@ class SysFS(list_ports_common.ListPortInfo):
 
             self.manufacturer = self.read_line(self.usb_device_path, 'manufacturer')
             self.product = self.read_line(self.usb_device_path, 'product')
-            self.interface = self.read_line(self.device_path, 'interface')
+            self.interface = self.read_line(self.usb_interface_path, 'interface')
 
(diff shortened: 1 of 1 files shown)


zsquareplusc:
can't reproduce with the composite device i have here (does not show an `interface` in any of the paths). Have you found some documentation on sys/.../interface?

johannesacco:
I have tested with CP2105, CP2108, TTL232RG-VIP. All of them show "None" as interface without this PR.

With this PR the interface property is correctly initialized to match the one in the USB Descriptor (e.g. "lsusb -v -d 10c4:ea70  | grep iInterface", "lsusb -v -d 10c4:ea71  | grep iInterface" and "lsusb -v -d 0403:6001  | grep iInterface" for above devices.

@zsquareplusc which devices did you try out with and without this PR?

zsquareplusc:
getting `None` on an STM that emulates 2 interfaces, getting duplicate text (same text for `product` and `interface`) on CP2104. but at least not `None`..

johannesacco:
Thanks for merging!

Issue #499, opened 2020-04-17
Status of the project


WisdomPill:
Dear @bufferoverflow and @zsquareplusc 

I am a long time user of the library, lately I started playing with asyncio and I was curios about pyserial with asyncio, I noticed there is a library (pyserial-asyncio) that depends on pyserial and it is made by the same community but the support is outdated for that library and this one as well. My question is: is this repository active? Does someone plan to keep supporting this amazing library? I would be glad to help and make some pull requests like dropping python 3.2 and 3.3 for the beginning and make pyserial more integrated with asyncio. I see many pull requests without an answer and many things were other people like me would be glad to help with.
Best regards, WisdomPill

WisdomPill:
@bufferoverflow or @zsquareplusc any news? Please let some people help with maintaining this library. It is used by a big part of the python community and it is showcased in many tutorials.

I am open to step in and help, I am interested also in the async version.

Can you please at least say something about the status of the project?

changyuheng:
Hi @WisdomPill, please check out [aioserial](https://github.com/changyuheng/aioserial). It's fully compatible with [pyserial](https://github.com/pyserial/pyserial).

jacobian91:
@bufferoverflow @zsquareplusc, It would be great to see development start again with this. I will second what @WisdomPill said about being a big part of the python community. I personally depend on it for work to talk to various hardware systems in a manufacturing environment.

kevindawson:
hi all, it's a long time since I did #380 
there was no reply then also
@WisdomPill it might be worth considering that the founders are _no more..._
I am happy to assist in making a **Quorum**

zsquareplusc:
pong

kevindawson:
ping @WisdomPill

WisdomPill:
Yes @kevindawson, as I said in the opening message, I am open to help since I think it's a shame that such a library is not maintained. It has been years since last release and the issues keep piling up. Also because it is the main choice when you look for serial communication with python. I think at least one or two people should actively be able to respond and review the proposed changes.

kevindawson:
@WisdomPill but for others to get involved - (spread the love) - we need a commit bit(s) - the backlog of **pull requests** needs to be addressed. the issues need to be collated - test updated - it's going to require a team - and most importantly an **active** dictator to led/arbitrate.
thoughts???

bufferoverflow:
@kevindawson @WisdomPill How many MR's have you reviewed here? On how many issues have you commented?

[...]

Issue #476, opened 2019-12-28
Travis-CI Check Failing (404 Not Found Error)


jacobian91:
The latest commit 3a6ae5a and a few of the latest pull requests show the Travis-CI checks failing on Python 3.2 & 3.3 for the same reason, **404 Not Found**.

```
3.2 is not installed; attempting download

Worker information

Build system information

Downloading archive: https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/16.04/x86_64/python-3.2.tar.bz2
$ curl -sSf -o python-3.2.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 404 Not Found
Unable to download 3.2 archive. The archive may not exist. Please consider a different version.
```

ksunden:
Looks like Travis removed old pythons.... We now use Azure pipelines anyway, AFAIK the Travis ones were never disabled but are not heeded by any of the devs.

WisdomPill:
I am making a pull request for dropping python 3.2 and 3.3, even 3.4 might be dropped but in my opinion only if necessary. I hope to get it merged and get the project active again. I already created an issue about that #499

WisdomPill:
I guess this can be closed now please @bufferoverflow or @zsquareplusc close this

PR #500, opened 2020-04-17
Dropping python 3.2 and 3.3

WisdomPill:
As previously mentioned in #499 and #476 this pull request is for dropping older python versions that are not available on travis anymore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ dist
 *.egg-info
 
 /MANIFEST
+
+.idea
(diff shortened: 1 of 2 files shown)


klanmiko:
I think this is simple enough to merge in ASAP

zsquareplusc:
yes makes sense, thank you

bufferoverflow:
> yes makes sense, thank you

@zsquareplusc I think you can merge this

Issue #437, opened 2019-06-04
Reusing exception instances stacks their tracebacks


dequis:
This was a fun one to debug, we have an exception handler that waits for a few seconds and calls the main function again (with exponential backoff and all the bells and whistles)

When it fails with exceptions from pyserial, the traceback starts getting longer every time it fails.

Here's some minimal reproduction steps:

```
$ python
Python 3.7.3 (default, Mar 26 2019, 21:43:19)
[GCC 8.2.1 20181127] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from serial import Serial
>>> a = Serial()
>>> a.write('')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dx/test/venv/lib/python3.7/site-packages/serial/serialposix.py", line 554, in write
    raise portNotOpenError
serial.serialutil.SerialException: Attempting to use a port that is not open
>>> a.write('')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dx/test/venv/lib/python3.7/site-packages/serial/serialposix.py", line 554, in write
    raise portNotOpenError
  File "<stdin>", line 1, in <module>
  File "/home/dx/test/venv/lib/python3.7/site-packages/serial/serialposix.py", line 554, in write
    raise portNotOpenError
serial.serialutil.SerialException: Attempting to use a port that is not open
>>> a.write('')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dx/test/venv/lib/python3.7/site-packages/serial/serialposix.py", line 554, in write
    raise portNotOpenError
  File "<stdin>", line 1, in <module>
  File "/home/dx/test/venv/lib/python3.7/site-packages/serial/serialposix.py", line 554, in write
    raise portNotOpenError
  File "<stdin>", line 1, in <module>
  File "/home/dx/test/venv/lib/python3.7/site-packages/serial/serialposix.py", line 554, in write
    raise portNotOpenError
serial.serialutil.SerialException: Attempting to use a port that is not open
```

With more complex call flows this looks even more confusing, with the traceback jumping from the exception point to outside the exception handler, with nothing connecting those pairs of frames.

I think this is due to the fact that portNotOpenError is an instance of an exception that exists at the module level, so reusing that instance will keep the traceback from the previous time it was used. It should probably be a class instead.

I can send a pull request with a fix for this later.

zsquareplusc:
[...]

PR #502, opened 2020-05-05
fix: stop reusing exception instances

j123b567:
writeTimeoutError and portNotOpenError ware precreated exception
instances. When rising these exceptions multiple times, stack trace
was appended to existing traces. This leads to two issues

 - no easy to debug real cause of the exception because of wrong
   stack trace
 - memory leak, because stack trace is never cleared. Raising
   portNotOpenError multiple times (understand e.g. one milion
   times, for long running script, it is perfectly possible)
   will cause out of memory

this change just assign specially crafted exception classes to
existing writeTimeoutError and portNotOpenError variables so
they will produce identical message but they are no longer
precreated.

New exceptions are derived from original SerialTimeoutException
and SerialException so existing user code should not break by this
change

closes: #437

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -97,8 +97,18 @@ class SerialTimeoutException(SerialException):
     """Write timeouts give an exception"""
 
 
-writeTimeoutError = SerialTimeoutException('Write timeout')
-portNotOpenError = SerialException('Attempting to use a port that is not open')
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thanks. I preferred to make instances in the `raise` calls instead of using exception classes. Therefore closing this w/o merging.

PR #506, opened 2020-05-26
Attempt to hot-reopen the USB device if it unexpectedly disappears/reappears

adamm:
Discovered an issue while testing PlatformIO related to stability of pyserial when using a faulty cable or an underpowered USB hub: pyserial would spontaneously crash mid-transfer:

```
Exception in thread rx:
  File "/Users/adamm/.platformio/penv/lib/python3.7/site-packages/serial/serialposix.py", line 550, in read
    buf = os.read(self.fd, size - len(read))
OSError: [Errno 6] Device not configured

During handling of the above exception, another exception occurred:

  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/adamm/.platformio/penv/lib/python3.7/site-packages/serial/tools/miniterm.py", line 447, in reader
    data = self.serial.read(self.serial.in_waiting or 1)
  File "/Users/adamm/.platformio/penv/lib/python3.7/site-packages/serial/serialposix.py", line 566, in read
serial.serialutil.SerialException: read failed: [Errno 6] Device not configured
```

Examining the issue further, I found simply closing and reopening the device allowed for an attempt to recover, and at least print a warning message to the user to change their cable, and/or power their USB hub if applicable.

pyserial: 3.4.1
OS: Mac OS 10.13.16 and 10.14.6
USB Serial: CP2102 USB to UART Bridge Controller
Device: ESP32

(NOTE, I couldn't find a good example of a STDERR warning, so tried to pick something obvious to the end-user as to not confuse it with legit serial output.)

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -559,10 +559,31 @@ class Serial(SerialBase, PlatformSpecific):
                         '(device disconnected or multiple access on port?)')
                 read.extend(buf)
             except OSError as e:
+                # Devices connecting through a poor USB/Serial cable or an
+                # underpowered USB hub can spontaneously disconnect from
+                # the OS, and promptly reconnect.  If this happens, attempt
+                # to reopen the device and continue the session.
+                if e.errno == errno.ENXIO:
+                    retryTimeout = Timeout(2)  # Allow for a short grace period
+                    sys.stderr.write(
+                        'pySerial WARNING: Serial device disappeared! '
+                        '...Trying to reconnect...\n')
+                    sys.stderr.write(
+                        'pySerial WARNING: If this problem is chronic '
+                        'replace your cable/powered USB HUB.\n')
+                    sys.stderr.flush()
+                    self.close()
+                    while self.fd == None:
+                        try:
+                            self.open()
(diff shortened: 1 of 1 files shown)


adamm:
The failing checks appear to be related to the travis-ci pipeline: Ubuntu and Python download errors <!>.

zsquareplusc:
Kind of interesting but it could also hide hardware problem from the user. The error could also be handled by the application that uses pySerial. I'm also not a fan of printing to stderr, it's probably not visible in many applications and others may prefer the logging module or something else.

Maybe a separate `class DisconnectedError(SerialException)` that would abstract the OS and make it easy for an application to handle would be useful. Especially if someone figures out how to detect this state on windows or other supported platforms too (should be possible for loop and rfc2217 too and fit nicely for those).

PR #514, opened 2020-07-11
Adding COBS encoding/decoding to the miniterm example script

jpaximadas:
I added a feature to the miniterm example that allows for COBS encoding.
The feature uses the cobs python library and requirements.txt has been updated.
Whether or not the miniterm is in COBS mode is expressed by the new Miniterm attribute ```self.cobs```.
Appropriate modifications have been made to the menu, help text, and argument parsing. (CTRL+C toggles the cobs mode and -c/--cobs are the flags).

The actual encoding/decoding is performed by the reader and writer threads:

### Reader thread modifications
If self.cobs is true, the script aggregates bytes in a buffer until a terminator (0x00) is received.
The script then COBS decodes the buffer. After decoding, the reader thread continues which prints the received data.
The cobs decode function can throw an error which is handled by printing a warning to the terminal along with the hex of the packet that failed to decode.
If cobs is toggled while the buffer is not empty, the reader continues without attempting to decode.

 ### Writer thread modifications
If self.cobs is true, the script aggregates bytes in a buffer until a newline is typed.
The script then COBS encodes the buffer. After encoding, the writer thread transmits the encoded data.
If cobs is toggled while frame bytes are being aggregated from the terminal, the reader will transmit without encoding the data.

If self.cobs is false, the script runs normally.


kurtmckee:
@zsquareplusc @projectgus Please close this PR. miniterm is a simple terminal, and COBS appears to be a niche use case.

projectgus:
Yes, agree this is out of scope.

@jpaximadas this is probably long in your past, but this could be a good rationale for a more feature-rich console terminal program.

PR #518, opened 2020-08-09
macOS Big Sur Support

JoelOsler:
Added macOS Big Sur support, as Big Sur removes access to Apple's Frameworks to the user, making os.path.isfile() fail.
os.path.isfile reads differently than direct file access.
The user doesn't have access to the Frameworks because of this -> https://www.reddit.com/r/MacOSBeta/comments/hfknpa/is_corefoundation_missing_for_everyone_on_big_sur/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ dist
 *.egg-info
 
 /MANIFEST
+
+.idea
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ language: python
 
 python:
   - 2.7
-  - 3.2
-  - 3.3
   - 3.4
   - 3.5
   - 3.6
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -734,3 +734,60 @@ Bugfixes (posix):
 - [#227] posix: ignore more blocking errors and EINTR, timeout only
   applies to blocking I/O
 - [#228] fix: port_publisher typo
+
+
+Version 3.5   2020-xx-xx
+------------------------
+New Features:
+
+- [#354] Make ListPortInfo hashable
+- [#411] Add a backend for Silicon Labs CP2110/4 HID-to-UART bridge. (depends on `hid` module)
+
+Improvements:
+
+- [#315] Use absolute import everywhere
+- [#351] win32: miniterm Working CMD.exe terminal using Windows 10 ANSI support
+- [#372] threaded: "write" returns byte count
+- [#400] Add bytesize and stopbits argument parser to tcp_serial_redirect
+- [#408] loop: add out_waiting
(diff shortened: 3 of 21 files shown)


JoelOsler:
I made this edit because I am on Big Sur, and I was attempting to use my ESP8266 NodeMCU 1.0 in Arduino IDE, but it was complaining. So I did a deep dive into the python code and learned about the new changes, and added what I learned into the code.

zsquareplusc:
Can someone confirm that this will not break support for older MacOS releases? I guess the paths to the libraries did not change, but to be sure...

ivankravets:
@zsquareplusc how about to move it under try/except? If 
```
iokit = ctypes.cdll.LoadLibrary(ctypes.util.find_library('IOKit'))
iokit = ctypes.cdll.LoadLibrary("/System/Library/Frameworks/IOKit.framework/Versions/Current/IOKit")
cf = ctypes.cdll.LoadLibrary(ctypes.util.find_library('CoreFoundation'))
```
does not work, then

```
iokit = ctypes.cdll.LoadLibrary("/System/Library/Frameworks/IOKit.framework/Versions/Current/IOKit")
cf = ctypes.cdll.LoadLibrary(ctypes.util.find_library('CoreFoundation'))
cf = ctypes.cdll.LoadLibrary("/System/Library/Frameworks/CoreFoundation.framework/Versions/Current/CoreFoundation")
```

In any case, I don't think that this PR is the best solution for this issue but it's good as a temporary workaround. We have over 1,000,000 developers for @PlatformIO and all macOS users will be affected :(

ivankravets:
It looks like an issue with CPython https://github.com/python/cpython/pull/21241

@JoelOsler could adapt your PR to try/except as I described above? It will work on built-in macOS Big Sur' Python 3 and with the future Python versions that have a fix.

Thanks!

zsquareplusc:
That PR (21241) has a test that suggests that IOKit can be in one of two locations, so a hardcoded path may not work on all versions of MacOS.

ivankravets:
Could we also iterate over 
```
        self.assertIn(find_lib('IOKit'),
                             '/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit')
                             ('/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit',
                             '/System/Library/Frameworks/IOKit.framework/IOKit'))
```
?

zsquareplusc:
looking at https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/FrameworkAnatomy.html i believe the short form of `/path/to/x.framework/x` should be sufficient and work across old and new versions of MacOS

zsquareplusc:
beta release is available, please see #528

PR #519, opened 2020-08-21
linux: Fix custom baud rate to not temporarily set 38400 baud rates

tim-nordell-nimbelink:
When using a custom baud rate on Linux these lines of code caused it
to push a 38400 baud rate setting down into the serial driver (even
if temporarily).  It would correct this moments later, but we can
simply set the BOTHER value at this point.  If other platforms have
a different default value for ispeed and ospeed when using a custom baud
rate, they can define BOTHER as well.


zsquareplusc:
Thanks.

PR #525, opened 2020-09-15
posix: VTIMESerial allows for inter-byte timeout

hitus:
Pass inter_byte_timeout to allow reads to return large blocks of data
until there is a gap between characters of 'inter_byte_timeout'
milliseconds or more.

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -834,7 +834,7 @@ class VTIMESerial(Serial):
         fcntl.fcntl(self.fd, fcntl.F_SETFL, 0)  # clear O_NONBLOCK
 
         if self._inter_byte_timeout is not None:
-            vmin = 1
+            vmin = 255
             vtime = int(self._inter_byte_timeout * 10)
         elif self._timeout is None:
             vmin = 1
@@ -867,11 +867,8 @@ class VTIMESerial(Serial):
         if not self.is_open:
             raise PortNotOpenError()
         read = bytearray()
-        while len(read) < size:
-            buf = os.read(self.fd, size - len(read))
-            if not buf:
-                break
-            read.extend(buf)
+        buf = os.read(self.fd, size - len(read))
+        read.extend(buf)
         return bytes(read)
 
     # hack to make hasattr return false


serial/serialposix.py
```diff
@@ -867,11 +867,8 @@ def read(self, size=1):
         if not self.is_open:
             raise PortNotOpenError()
         read = bytearray()
-        while len(read) < size:
-            buf = os.read(self.fd, size - len(read))
-            if not buf:
-                break
-            read.extend(buf)
+        buf = os.read(self.fd, size - len(read))
+        read.extend(buf)
```
zsquareplusc, line 871: Without the loop, could also simplify to `return os.read(...`

On the other hand, the loop could be kept to catch EAGAIN, EALREADY, EWOULDBLOCK, EINPROGRESS, EINTR like in the `read()` function of the default implementation.

hitus: Well, yes.  The bytearray could go away.

As for the loop: no.  Restarting the read will also restart timeout and by this change the timing.  I'm using pyserial for pymodbus and the read has to return within 3.5 character's time to distinguish between two packets.

Any serial application should be able to cope with a read returning zero bytes.  But what I did for pyserial 3.3 was to catch EAGAIN plus EINTR and ignore them (returning zero bytes immediately) to not have to handle that in the aplication.  Those exceptions will be delivered by the kernel when waiting is interrupted, which in case of a Modbus communication with a high probability would be either before the packet (same as overall timeout expired) or after it when VTIME is used to detect the end of the packet.

But the current state of VTIMESerial.read did not handle exceptions at all, so I was reluctant to add that.  And as I just detected: in 3.3 I've changed class Serial and not VTIMESerial.  I'm not to fond of two classes, why does VTIMESerial exist?

zsquareplusc: I've `VTIMESerial` as alternative implementation if for some strange reason, the default Serial class would not work. This one uses only the timeout functionality from the serial driver while `Serial` uses `select.select` to do the global timeout.
I once wanted to switch to `select.poll` just to find out that not all the fruity, Posixy OS implement that. So I kept those alternatives. They can be selected via the `alt://` URL handler.

PR #530, opened 2020-09-30
Fix exception for composite serial number search on Windows

jrowberg:
I tested the latest beta code on a composite device I'm developing with, and found that a malformed serial number string (which contains non-alphanumeric characters) causes the serial number search to raise an exception:

```
  File "serialtest.py", line 5, in <module>
    for port_info in serial.tools.list_ports.comports():
  File "C:\Python\WPy64-3830\python-3.8.3.amd64\lib\site-packages\serial\tools\list_ports_windows.py", line 412, in comports
    return list(iterate_comports())
  File "C:\Python\WPy64-3830\python-3.8.3.amd64\lib\site-packages\serial\tools\list_ports_windows.py", line 334, in iterate_comports
    info.serial_number = get_parent_serial_number(devinfo.DevInst, info.vid, info.pid)
  File "C:\Python\WPy64-3830\python-3.8.3.amd64\lib\site-packages\serial\tools\list_ports_windows.py", line 222, in get_parent_serial_number
    return get_parent_serial_number(devinst, child_vid, child_pid, depth + 1)
  File "C:\Python\WPy64-3830\python-3.8.3.amd64\lib\site-packages\serial\tools\list_ports_windows.py", line 197, in get_parent_serial_number
    vid = int(m.group(1), 16)
AttributeError: 'NoneType' object has no attribute 'group'
```

A quick search revealed why (and prompted me to change the serial number format of my device--I didn't know the spec said alphanumeric only). However, pyserial should fail gracefully here instead of crashing. I've modified the otherwise great traversal code so that it detects a failed regex match and ends the search early, instead returning (if available) the last identified serial value string from the previous iteration. No more crash, and now it's friendly enough to give a "best effort" value in a bad situation, even though the correct fix is to properly format your serial number string. Linux and macOS report non-alphanumeric serial content without issue, so I figured this was a reasonable approach.

### DIFF
--- a/serial/tools/list_ports_windows.py
+++ b/serial/tools/list_ports_windows.py
@@ -149,7 +149,7 @@ KEY_READ = 0x20019
 MAX_USB_DEVICE_TREE_TRAVERSAL_DEPTH = 5
 
 
-def get_parent_serial_number(child_devinst, child_vid, child_pid, depth=0):
+def get_parent_serial_number(child_devinst, child_vid, child_pid, depth=0, last_serial_number=None):
     """ Get the serial number of the parent of a device.
 
     Args:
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thank you.

PR #540, opened 2020-11-04
serialposix: Fix inconstent state after exception in open()

henzef:
If an exception occured during _update_dtr_state(), the Serial object
was left in an inconsistent state: is_open=True, but close() fails.
This patch fixes this by not setting is_open to True and by cleaning
up the state if an exception occurs during open()

I noticed this issue, because I would sometimes get

      File "mycode.py", line 159, in open
        ret = self._serial.open()
      File "/usr/local/lib/python3.7/dist-packages/serial/serialposix.py", line 286, in open
        self._update_dtr_state()
      File "/usr/local/lib/python3.7/dist-packages/serial/serialposix.py", line 634, in _update_dtr_state
        fcntl.ioctl(self.fd, TIOCMBIS, TIOCM_DTR_str)
    OSError: [Errno 5] Input/output error

on pyserial 3.4, which is likely related to replugging the device with unlucky timing.
A 3.4 version of this patch is available at: https://gist.github.com/henzef/0bb02041a43365d93f5bd6265313de72

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -325,35 +325,53 @@ class Serial(SerialBase, PlatformSpecific):
             raise SerialException(msg.errno, "could not open port {}: {}".format(self._port, msg))
         #~ fcntl.fcntl(self.fd, fcntl.F_SETFL, 0)  # set blocking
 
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thank you

PR #546, opened 2020-11-18
fix issue 545 - getting USB info on BigSur/AppleSilicon

chairman888:
added call to get IOUSBHostDevice before falling back to
IOUSBDevice (which has been removed on Big Sur on Apple Silicon)

Thanks to @oskay for this patch.

### DIFF
--- a/serial/tools/list_ports_osx.py
+++ b/serial/tools/list_ports_osx.py
@@ -270,7 +270,11 @@ def comports(include_links=False):
         if device:
             info = list_ports_common.ListPortInfo(device)
             # If the serial port is implemented by IOUSBDevice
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thank you, both.

RschDev:
Is this a dead project or was a solution ever found?

oskay:
> Is this a dead project or was a solution ever found?

Even if a solution were not found, that would not make this a "dead project". This solution was merged, meaning that it was accepted and is now part of the release.

sergioherrera7:
Using Version=3.5 keeps showing some devices as "n/a" in my M1 python3 setup

Issue #467, opened 2019-11-27
spy://?raw broken at Python3


Harvie:
`serial.write("hello")` gives error:
`unicode strings are not supported, please encode to bytes: 'hello'`
so i encode to bytes like this: `serial.write("hello".encode())`, which gives me error:
`write() argument must be str, not bytes`.

But as far as i know all str's in python3 are unicode.

When i provide string, it says me to provide bytes. But when i provide bytes, it says me to provide string. How can i get out from this catch-22 ?

Harvie:
So i've tracked the problem down bit further. It happens only with `spy://` in `?raw` mode.

This works:

```
import serial
s = serial.serial_for_url("spy:///dev/ttyUSB3", 115200)
s.write("hello".encode())
s.close()
```
This gives me an error:

```
import serial
s = serial.serial_for_url("spy:///dev/ttyUSB3?raw", 115200)
s.write("hello".encode())
s.close()
```

rkandilarov:
Just to confirm that I got in the very same problem - 'raw' in 'spy://' and dig it to the very same code, patched it, and then decided to see if anybody was done it before I sent a pull request :).... and @Harvie  had done it.

So the proposed solution works, and will be great to see it in the next release!

Other solution could be reopening the file in binary mode `output = open(values[0], 'wb')` instead of only 'w' when 'raw' mode.

Regards!

acerv:
I'm also stucking with the same problem. It would be great to fix it! :)
Thanks

PR #553, opened 2020-12-28
Fix raw spy formating on python3 #467 (tested on both py2 and py3)

Harvie:
Fix raw spy formating on python3 #467 (tested on both py2 and py3)


kurtmckee (approved):
@zsquareplusc @dpgeorge Please merge this PR. It correctly decodes `bytes` to `str` which matches the types expected by the output writers (like `sys.stderr.write()`).

kurtmckee:
Closes #467.

This may need to be closed manually because it is not auto-linked to this PR.

PR #566, opened 2021-02-24
fix interface names on macOS in list_ports.comports

Neradoc:
On macOS Catalina and Big Sur, when connecting a USB device with 2 serial ports, both got associated with the name of the first one. For example running this with a microcontroller exposing two serial port returns the same name for each interface, which it shouldn't:
```py
>>> import serial.tools.list_ports
>>> for port in serial.tools.list_ports.comports():
...     print(port.device,port.interface)
```
```
/dev/cu.Bluetooth-Incoming-Port None
/dev/cu.usbmodem144443121 CircuitPython CDC data
/dev/cu.usbmodem144443123 CircuitPython CDC data
```
That last line should read "CircuitPython CDC2 data"
The interfaces listed by `scan_interfaces()` get associated the same locationID for all ports on the same device, so they all get the name of the first one.

This fix gets the "USB Interface Name" from the current device's nearest `IOUSBInterface` parent before falling back to the scan's list. I don't know enough about IOKit to know if using the scan is still necessary.
```
/dev/cu.Bluetooth-Incoming-Port None
/dev/cu.usbmodem144443123 CircuitPython CDC2 data
/dev/cu.usbmodem144443121 CircuitPython CDC data
```

### DIFF
--- a/serial/tools/list_ports_osx.py
+++ b/serial/tools/list_ports_osx.py
@@ -270,6 +270,8 @@ def comports(include_links=False):
         device = get_string_property(service, "IOCalloutDevice")
         if device:
             info = list_ports_common.ListPortInfo(device)
+            # find the serial interface associated with this device
+            serial_interface = GetParentDeviceByType(service, "IOUSBInterface")
             # If the serial port is implemented by IOUSBDevice
             # NOTE IOUSBDevice was deprecated as of 10.11 and finally on Apple Silicon
             # devices has been completely removed.  Thanks to @oskay for this patch.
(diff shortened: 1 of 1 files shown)


Neradoc:
Updated for Apple Silicon
- `IOUSBInterface` becomes `IOUSBHostInterface`
- `USB Interface Name` removed for `kUSBString`

Updated to look for the interface name outside of the hierarchy on 10.12 and 10.13, using a rewritten scan function.

rdagher:
@dhalbert @aivarannamaa Any chance this getting merge to master any time soon?
I tested it on my macbook pro arm m3 and the fix suggested by @Neradoc fixes the issue having info.interface fied set None using the code in the master branch.
Tested with a silabs CP2105 device.

Current results from latest release
image

Results applying this PR fix
image

rdagher:
@zsquareplusc, @dpgeorge is this supposed to be solved with latest release?
Did a quick test witha a silabs CP2105, and it seems not: description string is always the same.


Thanks !

dpgeorge:
@rdagher this fix has not been merged yet.

PR #578, opened 2021-04-28
Add support for cygwin

### LINKED ISSUE #379: Cygwin:  re-open port on resetted serial connection
I am trying to use the pygatt module with python 2.7.14 under cygwin to communicate with a BLE dongle.
During initialization of the API, the constructor of the API open serial connection to the dongle. And send a reset command via the serial interface, flush and close the serial connection. Then the API reopens the serial connection. And at this point it hit an exception,

>   File "/usr/lib/python2.7/site-packages/serial/serialutil.py", line 282, in __init__
>     self.open()
>   File "/usr/lib/python2.7/site-packages/serial/serialposix.py", line 296, in open
>     self._reconfigurePort()
>   File "/usr/lib/python2.7/site-packages/serial/serialposix.py", line 325, in _reconfigurePort
>     raise SerialException("Could not configure port: %s" % msg)
> SerialException: Could not configure port: (22, 'Invalid argument')

The worst part is this effect seems to live across process, when I ran the script again, exception will happen with the first open which is done in the constructor of Serial. 

I try running the exact same code under Anaconda python and it work without a problem. Then tried to dig deeper, re implement the way that pygatt open/closes/reconnect serial port under Cygwin, but came up empty and everything seems to work fine. 

Then I commented out the serial command the pygatt module send to reset the dongle and it worked.

I suspect the when the dongle reset, Serial process that was connected to it somehow locked up when running in Cygwin python otherwise I can't explain issue sustain across processes. Is there anyway to clean up the serial objects completely along with any process it is associated with?

nhatkhai:
Add support for cygwin may also fix #379
Ran though run_all_tests.py on cygwin, and on windows. Their result are identical.

### DIFF
--- a/serial/__init__.py
+++ b/serial/__init__.py
@@ -22,6 +22,8 @@ VERSION = __version__
 # pylint: disable=wrong-import-position
 if sys.platform == 'cli':
     from serial.serialcli import Serial
+elif sys.platform == 'cygwin':
+  from serial.serialcygwin import Serial
 else:
     import os
     # chose an implementation, depending on os
--- a/serial/cygwin.py
+++ b/serial/cygwin.py
@@ -1,0 +1,395 @@
+#! python
+#
+# Constants and types for use with Windows API, used by serialwin32.py
+#
+# This file is part of pySerial. https://github.com/pyserial/pyserial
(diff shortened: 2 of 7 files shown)


nhatkhai:
Feedback or merge this please p

PR #382, opened 2018-09-10
serial/rfc2217: add timeout.setter for rfc2217

Emantor:
Add a new setter method for the timeout property which does not invoke
the port reconfiguration.

Signed-off-by: Rouven Czerwinski <r.czerwinski@pengutronix.de>

PR #579, opened 2021-04-30
refactor checks for valid timeouts

ukleinek:
This unifies the error paths for the two checks done on new timeout
values. The resulting checking is slightly stricter, but I don't expect
a real impact in practice. (For example complex numbers and timeouts of
type decimal.Decimal are refused now with the former resulting in
problems afterwards.)

This was suggested by penguinolog in a review comment to
https://github.com/pyserial/pyserial/pull/382.

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -361,11 +361,7 @@ class SerialBase(io.RawIOBase):
     def timeout(self, timeout):
         """Change timeout setting."""
         if timeout is not None:
-            try:
-                timeout + 1     # test if it's a number, will throw a TypeError if not...
-            except TypeError:
-                raise ValueError("Not a valid timeout: {!r}".format(timeout))
-            if timeout < 0:
+            if not isinstance(timeout, (int, float)) or timeout < 0:
                 raise ValueError("Not a valid timeout: {!r}".format(timeout))
         self._timeout = timeout
         if self.is_open:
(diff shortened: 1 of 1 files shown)


NickCrews (changes requested):
My two cents: In general this is a great idea to simplify this logic. I think I might prefer if instead of just simplifying each instance, we actually moved the logic to a separate `_is_valid_timeout(timeout)` static method. With your change, there is still the danger that one of the call sites will get updated but not the others, leaving them out of sync. However, I think they always should be in sync. Also, if contained happily in its own function, I don't think it's worth it to change the actual implementation, that ugliness can just stay as is.

In fact, looking at this more, I think this factored out method should contain _all_ the logic, eg
```
@staticmethod
def _assert_valid_timeout(timeout):
    if timeout is None:
        return
    e = ValueError("Not a valid timeout: {!r}".format(timeout))
     if timeout < 0:
        raise e    
    try:
        timeout + 1  # test if it's a number, will throw a TypeError if not...
    except TypeError:
        raise e
```
It also looks like these setter functions don't have any tests, they probably should get some simple ones to verify this change doesn't mess anything up. @ukleinek If you don't want to take on all these tasks, I can help as well.

Also, as I looked at this existing logic, there may be a bug/weird feature: The `if timeout < 0` check can raise a `TypeError` if timeout can't be compared to 0, but _that_ TypeError is never caught, so that TypeError bubbles up to the caller. This is out of scope of this PR though I think, if we even want to change it.

dpgeorge:
> I think I might prefer if instead of just simplifying each instance, we actually moved the logic to a separate `_is_valid_timeout(timeout)` static method

Yes, I think that's a good idea.  Then there's only one place to updated if it needs updating, and it reduces code duplication.

My suggestion would be to make a PR just to do the refactoring of the existing code, without changing the logic of how it works.   That will be relatively uncontroversial.  Then separately we can consider if the test should be changed (because currently this PR is a breaking change for the API).

PR #581, opened 2021-05-08
Add a note that the first argument to read_until was called `terminator` in previous versions

jonathangjertsen:


zsquareplusc:
Thanks

PR #584, opened 2021-05-14
Fix misc. documentation and source comment typos

luzpaz:
Found via `codespell -q 3 -L ba,ser,wont`  
First identified from a downstream project that uses the pyserial lib


zsquareplusc:
thank you

PR #589, opened 2021-06-15
Fix #588

### LINKED ISSUE #588: Allow read_until only bytes size
Allow the `read_until(foo)` to terminate read only by package size by set `expected=None`.

This is necessary to some device that base it answer on a binary format, the `LF` value can collide with some converted value (as some byte of IEEE 754 or some checksum).

hildogjr:
Allow finish the read only by package size.

I need this to capture the package off devices that answer in binary format and not ascii/human-readable. I intend to use `read_until(expect=None, size=8)` after I send a command to this device.

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -653,17 +653,19 @@ class SerialBase(io.RawIOBase):
 
     def read_until(self, expected=LF, size=None):
         """\
-        Read until an expected sequence is found (line feed by default), the size
+        Read until an expected sequence is found ('\n' by default), the size
         is exceeded or until timeout occurs.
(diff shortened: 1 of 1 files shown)


hildogjr:
@Someone (@zsquareplusc ?) could check it? I am already using and aims to go to PyPI for better install into my production ambience.

zsquareplusc:
but if you want to finish reading by timeout or size, the normal `Serial.read()` is providing this.

e.g. opening the port with `s = Serial(... timeout=10)` then `s.read(8)` will read 0..8 bytes from the port depending on what happens first the timeout (0..7 bytes returned) or 8 (if 8 or more bytes are sent to the PC).

PR #591, opened 2021-06-24
Add type stubs

9999years:
- These type stubs don't include any deprecated methods -- this will make code using deprecated methods incompatible with these stubs, but that's OK -- they're just renames, and we don't want autocomplete to suggest deprecated methods
- There's no stubs for the following modules (yet):
  - [ ] `rfc2217`
  - [ ] `rs485`
  - [ ] `threaded`
  - [ ] `urlhandler.*`

### DIFF
--- a/serial/__init__.pyi
+++ b/serial/__init__.pyi
@@ -1,0 +1,23 @@
+from typing import Any, List
+import sys
+
+from .serialutil import *
+
+protocol_handler_packages: List[str]
+
+__version__: str
+VERSION: str
+
+if sys.platform == "cli":
+    from .serialcli import Serial
+elif sys.platform == "win32":
+    from .serialwin32 import Serial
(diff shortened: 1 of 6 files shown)


9999years:
I'm unsubscribing from this thread as I no longer use pyserial. Feel free to merge these but I won't be modifying them in the future.

kurtmckee:
@zsquareplusc @andrewleech Please close this PR without merging.

Type annotations are valuable, but this PR introduces `.pyi` files instead of inline type annotations, which is an administrative burden to maintain. In addition, the codebase isn't yet running mypy to validate type annotations, so discrepancies and errors will be invisible except to downstream users.

I'm working to type-annotate the code base while expanding testing, and will add mypy testing in a final push later.

For now, please close this PR.

PR #602, opened 2021-08-18
Spellcheck

erikbgithub:
Really enjoying the changes `pyserial` made through the last decade. Kudos!

Wanted to make a small contribution myself, so I made a little spellcheck change. And while I'm not a native speaker, this way it sounds clearer to me. "as expected" sounds to me like "similar or equal to what you should expect", but here it means that you might expect 800 bytes and just receive 300.


zsquareplusc:
thank you

PR #604, opened 2021-08-28
make spy-`Serial` object with file sink pickelable

hojo0590:
Sometime you want to have your Serial connection in a process of its own. As a `Serial()` instance with an opened COM port can't be pickled on Windows you need to start the process and pass the `Serial()` instance into with its COM port being closed. This is already possible today.  
The `spy://` url-magic coming with `pyserial`'s `serial.serial_for_url()` offers so many benefits of which some you just can't use with this multiprocessed approach. One of these is writing all communication to a file. The `spy://` url-magic internally immediately opens the file to be written to - even if you pass `serial.serial_for_url(..., do_not_open=True)`. I aim to change this with this PR.  

I intentionally opened this as draft to have a discussion if that logic is useful to others, wanted and if it should be extended further to loggers and `sys.stderr` - allthough those don't always behave very well with multiprocessing anyhow. If this is wanted, some (minor) breaking changes of the url-magic might only be omitted with hacky solutions.

### DIFF
--- a/serial/urlhandler/protocol_spy.py
+++ b/serial/urlhandler/protocol_spy.py
@@ -27,7 +27,7 @@ import sys
 import time
 
 import serial
-from serial.serialutil import  to_bytes
+from serial.serialutil import to_bytes
 
 try:
     import urlparse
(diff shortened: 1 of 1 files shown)


hojo0590:
@zsquareplusc you have been the last one to work at exactly that corner of your library. I'd be very happy if you'd have a look at my proposed changes and to learn if this is appreciated and wanted and what'd be needed to be done in order to proceed.

PR #606, opened 2021-09-02
Fix duplicate comports listed with list_ports(include_links=True)

madislutter:
In case there are any symlinks which name matches any of the previously
checked globs then list_ports(include_links=True) will report them twice.
For example if there's a symlink named /dev/ttyUSB_mylink then it would
appear twice.

The problem is fixed by storing the found paths in a set instead of a list.
Documentation at https://pythonhosted.org/pyserial/tools.html#module-serial.tools.list_ports
already states that "Items are returned in no particular order" so there's
no need to do extra sorting before returning the results.

The changes have been tested on Ubuntu (using list_ports_linux.py).

### DIFF
--- a/serial/tools/list_ports_linux.py
+++ b/serial/tools/list_ports_linux.py
@@ -89,15 +89,16 @@ class SysFS(list_ports_common.ListPortInfo):
 
 
 def comports(include_links=False):
-    devices = glob.glob('/dev/ttyS*')           # built-in serial ports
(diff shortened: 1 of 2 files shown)


zsquareplusc:
ah, i see... that happens when the symlink name has one of the common prefixes and its therefore already included by the glob call. i always used totally different names, so not having this problem.

Issue #617, opened 2021-10-29
UnboundLocalError when using PosixPollSerial with timeout


Lukanite:
When using PosixPollSerial, with a timeout, an UnboundLocalError is thrown when the timeout expires without completely fulfilling the read request:
```
>>> ser = serial.serial_for_url("alt:///dev/ttyACM0?class=PosixPollSerial", 115200, timeout=10)
>>> x = ser.read(1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/dist-packages/serial/serialposix.py", line 830, in read
    if fd == self.pipe_abort_read_r:
UnboundLocalError: local variable 'fd' referenced before assignment
```

codewiz:
This bug is still present in master.

PR #618 by @Lukanite looks correct to me. Could a maintainer please review it?

PR #618, opened 2021-10-29
Fix UnboundLocalError on timeout

Lukanite:
On timeout, fd is no longer unbound for the check on line 838. Causes an os.read() still to be called, but it seems to be harmless on my machine. Addresses issue #617.

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -825,6 +825,7 @@ class PosixPollSerial(Serial):
         poll.register(self.pipe_abort_read_r, select.POLLIN | select.POLLERR | select.POLLHUP | select.POLLNVAL)
         if size > 0:
             while len(read) < size:
+                fd = None #In case poll.poll() returns nothing
                 # print "\tread(): size",size, "have", len(read)    #debug
                 # wait until device becomes ready to read (or something fails)
                 for fd, event in poll.poll(None if timeout.is_infinite else (timeout.time_left() * 1000)):


kurtmckee (commented):
@zsquareplusc @andrewleech Please merge this PR.

It fixes a crash that can occur if a `for` loop never executes and thus never creates the name `fd` in the local scope.

Code demonstrating the effect when `poll.poll()` doesn't iterate even once:

```python
for fd, event in ():
    pass

fd  # crash
```

dpgeorge:
But with this fix, it still calls `os.read(self.fd, ...)` even on timeout, in which case that read could block (this is mentioned above).

So I think a different fix is needed, either check for the timeout right after the for loop, or put all the fd-related code in the for loop (IMO that's the right fix):
```py
        for fd, event in poll.poll(None if timeout.is_infinite else (timeout.time_left() * 1000)):
            if fd == self.pipe_abort_read_r:
                os.read(fd, 1000)
                abort = True
            if event & (select.POLLERR | select.POLLHUP | select.POLLNVAL):
                raise SerialException('device reports error (poll)')
            buf = os.read(fd, size - len(read))
            read.extend(buf)
        if abort:
            break
        if timeout.expired() ...
```

kurtmckee:
@dpgeorge For expedience, I'll apply this suggestion to a new branch so that @Lukanite has credit for this work, but the outcome will be the fix you've suggested. I'll open a PR shortly for that, at which point this PR can be closed.

kurtmckee:
@zsquareplusc @dpgeorge Please close this PR. It was merged with an additional change in #820.

dpgeorge:
Merged in 991383aacfcbadf26cf7f17000651b063c0c10df

PR #627, opened 2021-12-04
Docs: \n was not in inline mode, causing the backslash to be escaped

spmvg:
Hi devs, thanks for the work on pyserial! Miniscule fix in the docs. `\n` was not in inline mode, causing the backslash to be escaped: https://pyserial.readthedocs.io/en/latest/shortintro.html#readline. Fixed by adding backticks:

image


zsquareplusc:
thank you

PR #638, opened 2022-02-13
Close opened file handler on closing the serial port

rkandilarov:
When using the  spy://___URI___?file=__SOME_FILE___&raw format for serial port the __SOME_FILE___ is opened and its file handler is not closed when the serial port is closed. This patch fixes this. By default the opened stream is 'stderr' which cannot be closed, so check for its existence is made.


rkandilarov:
Now I see that this is also fixed in other proposal, that do extends the spy funcionality in https://github.com/pyserial/pyserial/pull/604 .

PR #647, opened 2022-04-24
/serial/tools/list_ports_linux.py: add support for wwan* port

zixi11:
When the ports of the modem are wwan0at* and wwan0mbim*, comports
cannot list them. I made some changes to make it recognize it.


vanmierlo:
+    devices.update(glob.glob("/dev/wwan0*"))    # Intel pcie driver port(modem device)
I believe this part is better handled by patch #709

zixi11:
For  find modem, I don't have any question.

But  I want to  recognize the model vendor when we list the modems.

PR #649, opened 2022-04-28
Add '--data' and '--stop' options to miniterm

clvz:
Hi,

I've added two options to the CLI interface that makes it easier for me to use miniterm when dealing with old hardware.


zsquareplusc:
This is of course useful, thank you

PR #652, opened 2022-05-31
tools/miniterm: add swap LF for CR option

alefminus:
I have a terminal application that needs to get LF only for EOL but sends CR+LF, so if I use the LF option I get double spaced lines in the output. This simple addition fixes that.

### DIFF
--- a/serial/tools/miniterm.py
+++ b/serial/tools/miniterm.py
@@ -272,6 +272,11 @@ class LF(Transform):
     """ENTER sends LF"""
 
 
+class CRTX(Transform):
+    """ENTER sends CR, but receive CR+LF"""
+    def tx(self, text):
+        return text.replace('\n', '\r')
+
 class NoTerminal(Transform):
     """remove typical terminal control codes from input"""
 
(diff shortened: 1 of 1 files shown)


serial/tools/miniterm.py
```diff
@@ -288,6 +294,20 @@ def rx(self, text):
...
+
+    def __init__(self, *args, **kw):
+        super().__init__(*args, **kw)
+        self._start = datetime.now()
+
+    def rx(self, text):
+        ts = (datetime.now() - self._start).total_seconds()
+        return f'{ts:9.4f}: {text}'
+
+    echo = rx
```
kurtmckee: This change is unrelated to this PR.

alefminus: Will remove from PR

serial/tools/miniterm.py
```diff
@@ -272,6 +273,11 @@ class LF(Transform):
     """ENTER sends LF"""
 
 
+class LFTX(Transform):
+    """ENTER sends LF, but receive CR+LF"""
+    def tx(self, text):
+        return text.replace('\n', '\r')
```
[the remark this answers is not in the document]
alefminus: My mistake, I got them confused. Renaming to CRTX / crtx

Issue #655, opened 2022-06-23
List open serial ports *from python script*, not from the command line


beyarkay:
So I've got a serial device, however the serial port changes most times I reconnect it. 
What I want to do is to list all serial ports, iterate over them until I find the correct one, and then connect to that one.
I know that I can list all ports from the commandline like:
```sh
$ python -m serial.tools.list_ports
/dev/cu.YYYY
/dev/cu.XXXX
ZZZZ ports found
```

I want to do that from a python, but this doesn't work:
```python
>>> import serial
>>> for port in serial.tools.list_ports:
...     print(f'Current port: {port}')
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'serial' has no attribute 'tools'
```

What can I do? I really don't want to execute `python -m serial.tools.list_ports` as an external shell command, as it just seems silly since the library already has an API for python scripts.

Thanks!

b4sen:
`list_ports` is a python module, so you can't really iterate over it.

if you look at the code in `serial/tools/list_ports` you can see the `if __name__ == '__main__':` at the end. This is the "pythonic" way of defining an entrypoint (e.g. `int main(int argc, char* argv[])` in c). The `-m` argument after python says you would like to execute a python `module`. 

so the command you ran in your terminal will call the `main()` function in the `list_ports.py`, which calls the `comports()` function. 

in order to list the ports from python, you can do the following:

```python
from serial.tools.list_ports import comports

for port in comports():
    print(port)
```
I hope this helps.

beyarkay:
Ah okay thanks. Any complaints if I submit a pull request to the docs describing how to list all available ports?

PR #658, opened 2022-06-29
Add example about listing ports via python (not via cmd)

beyarkay:
This pull request resolves issue https://github.com/pyserial/pyserial/issues/655. It is only a documentation change, and adds one example describing how to use pySerial to list all available ports via python, and then adds references to this example in appropriate places where the alternative `python -m serial.tools.list_ports` is listed.

Hope this is useful! let me know if you'd like any changes.

### DIFF
--- a/documentation/examples.rst
+++ b/documentation/examples.rst
@@ -5,6 +5,18 @@
 ==========
 
 
+List all available serial ports
+===============================
+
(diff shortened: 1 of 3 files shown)


beyarkay:
Hi, is this repository still monitored?

PR #659, opened 2022-07-12
Allow passing default end-of-line (eol) terminator to miniterm.main()

jelledevleeschouwer:
Extend miniterm.main() with default_eol='CRLF' argument


zsquareplusc:
Thank you

PR #661, opened 2022-07-29
support ironpython new version

sensokame:
in the new versions of ironpython (3.4), some changes were needed

[platform detection method](https://github.com/IronLanguages/ironpython3/blob/master/Documentation/upgrading-from-ipy2.md#checking-for-ironpython)

[getting byte array from result](https://github.com/pyserial/pyserial/blob/0e7634747568547b8a7f9fd0c48ed74f16af4b23/serial/serialcli.py#L22)

### DIFF
--- a/serial/__init__.py
+++ b/serial/__init__.py
@@ -20,7 +20,7 @@ __version__ = '3.5'
 VERSION = __version__
 
 # pylint: disable=wrong-import-position
-if sys.platform == 'cli':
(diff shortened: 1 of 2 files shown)


kurtmckee:
@zsquareplusc @andrewleech Please close this PR.

Python 3.4 is not a supported language version, and IronPython is no longer a target platform.

PR #663, opened 2022-08-04
Allow for longer location Strings.

enCHineer:
When using a Composite device connected to a USB-Hub (e.g. HP-Docking Station) the location string is longer than the previous buffer of 250. This then fails to extract the correct string. Doubling the buffer size got it to work as expected!


gudnimg:
Seeing the same issue here. Stumbled on the same solution as in this PR 👍 Works well with 2x buffer size.

I'm using Windows 11 for what its worth.

zsquareplusc:
Thank you

PR #665, opened 2022-08-25
Add detection for Haiku's serial ports

OscarL:
Now that PySerial works OK (without patches. relying on serialposix.py) on the nightly versions of Haiku (see https://haiku-os.org), make sure its ports gets detected properly.


davet2001 (approved):
Looks good to me!

PR #637, opened 2022-02-02
list_ports: store USB product string descriptor

mkarhumaa:
Store the USB product string descriptor, also known as
"Bus reported device description" in Windows lingo, as ListPortInfo's
product attribute also on Windows. This functionality makes
pyserial to behave similarly on Linux and Windows.

Previuosly the USB product string descriptor was not used in
list_ports_windows at all.

This is slightly modified version of https://github.com/pyserial/pyserial/pull/338.

PR #687, opened 2023-03-10
list_ports: store USB product string descriptor

jupe:
Store the USB product string descriptor, also known as "Bus reported device description" in Windows lingo, as ListPortInfo's product attribute also on Windows. This functionality makes pyserial to behave similarly on Linux and Windows.

Previuosly the USB product string descriptor was not used in list_ports_windows at all.


jupe:
duplicate with https://github.com/pyserial/pyserial/pull/637

Issue #699, opened 2023-04-06
list_ports does not include serial ports by id


davet2001:
### Summary

When I use `list_ports` to provide a list of serial ports to a user in linux, they are only shown ports directly under `/dev/*` (excluding sudirectories).  This can result in multiple ports such being displayed e.g. /dev/ttyUSB0, /dev/ttyUSB1.  

This is a problem for two reasons:
- There is no obvious indication which port is which - often resulting in user trial and error.
- If the machine is restarted with USB adapters disconnected or in a different arrangement, the mapping of USB0,1 etc can change.

The solution would be to include serial ports by ID in the list.

### Steps to reproduce
What I did:
```
$ python3 -m serial.tools.list_ports -v -s
```
What I expected to ### happen:
```
$ python3 -m serial.tools.list_ports -v -s
/dev/serial/by-id/usb-1a86_USB2.0-Serial-if00-port0
    desc: n/a
    hwid: LINK=/dev/ttyUSB0 LINK=/dev/ttyUSB0
/dev/ttyUSB0        
    desc: USB2.0-Serial
    hwid: USB VID:PID=1A86:7523 LOCATION=1-4
2 ports found
```
What actually happened:
```
/dev/ttyUSB0        
    desc: USB2.0-Serial
    hwid: USB VID:PID=1A86:7523 LOCATION=1-4
1 ports found
```

I am working on a PR for this.

SputnikTheMighty:
Agreed, would also be really nice to have the same thing for `dev/serial/by-path/` as well

PR #700, opened 2023-04-06
Include /dev/serial/by-id/* devices in links

davet2001:
When providing a user with an option to select the serial port in linux, the links under `/dev/serial/by-id/*` are not shown.

Identifiers such as `/dev/ttyUSB1` etc are not reliable through reboots or reconfiguration, see #699.

This PR adds an the additional search path `/dev/serial/by-id/*`, so that links to ports can be shown to the user by their serial number identifier as well as the `ttyUSBx` style name.

I have run the tests via `test/run_all_tests.py`, and all passed.

### DIFF
--- a/serial/tools/list_ports_common.py
+++ b/serial/tools/list_ports_common.py
@@ -109,7 +109,7 @@ def list_links(devices):
     listed in devices.
     """
     links = []
-    for device in glob.glob('/dev/*'):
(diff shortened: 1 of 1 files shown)


MinePlugins:
+1

davet2001:
@zsquareplusc Please could you advise whether this is still actively being maintained?  I see no commits for ~2y but definitely people are still using this.

PR #702, opened 2023-04-20
change the position of The open()

P-fanzr:
change the position of The open()

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -239,9 +239,9 @@ class SerialBase(io.RawIOBase):
             self.inter_byte_timeout = kwargs.pop('interCharTimeout')
         if kwargs:
             raise ValueError('unexpected keyword arguments: {!r}'.format(kwargs))
-
-        if port is not None:
-            self.open()
+        # self.open() shouldn't be here, The open function in rfc2217.py ,so it should be in Serial
+        # if port is not None:
+        #     self.open()
 
     #  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -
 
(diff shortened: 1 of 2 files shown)


serial/serialutil.py
```diff
@@ -239,9 +239,9 @@ def __init__(self,
             self.inter_byte_timeout = kwargs.pop('interCharTimeout')
         if kwargs:
             raise ValueError('unexpected keyword arguments: {!r}'.format(kwargs))
-
-        if port is not None:
-            self.open()
+        # self.open() shouldn't be here, The open function in rfc2217.py ,so it should be in Serial
```
JonathonReinhart, line 242: You seem to misunderstand how inheritence works in object-oriented programming.

- `serial.rfc2217.Serial.__init__`  calls `super(Serial, self).__init__()` which results in this function, `SerialBase.__init__`.
- `SerialBase.__init__()` calls `self.open()`
  - `self` still refers to a `serial.rfc2217.Serial` object, so `open` (due to method resolution order) refers to `serial.rfc2217.Serial.open()` (the function you want it to call).

If you remove this call to `self.open()`, how would any of the other classes derived from `SerialBase` actually open their serial ports?

JonathonReinhart:
Sorry, but this PR is completely wrong.

It's not clear what problem you're trying to solve.

serial/serialutil.py
P-fanzr: I think you can review the code. In fact, this method is just only used in The son class,OK?

P-fanzr: Besides, if the parent class is instantiated and the self.open method is called, it will result in an error

P-fanzr:
bitch  son of bitch  fuck  bitch

PR #721, opened 2023-10-31
 Use unittest.TestLoader.loadTestsFromModule() to load modules

befeleme:
The old unittest.findTestCases() has been removed from Python 3.13
See: https://docs.python.org/dev/whatsnew/3.13.html#unittest

We've encountered this during the early integration of Python 3.13 in Fedora Linux, this has unblocked the package build.

### DIFF
--- a/test/run_all_tests.py
+++ b/test/run_all_tests.py
@@ -26,6 +26,7 @@ if len(sys.argv) > 1:
 
 # find files and the tests in them
 mainsuite = unittest.TestSuite()
+loader = unittest.TestLoader()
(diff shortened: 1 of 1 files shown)


kurtmckee:
@zsquareplusc @dpgeorge Please close this PR.

#757 addressed this same issue and recently merged.

dpgeorge:
Hmm, this is probably marginally better than the merged #757, because it only creates one `TestLoader` instance.

Well, I guess the issue is fixed so can close this.

PR #725, opened 2023-12-19
win32: Get the string description of the usb serial port via DeviceIoControl.

chinaheyu:
This pull request helps to get the serial number, product string and manufacturer string of the usb serial port correctly on windows.

https://lihashgnis.blogspot.com/2018/07/getting-descriptors-from-usb-device.html?m=1
https://stackoverflow.com/questions/28007468/how-do-i-obtain-usb-device-descriptor-given-a-device-path
https://stackoverflow.com/questions/50701189/how-to-get-usb-device-descriptor-by-doing-deviceiocontrol-directly-on-the-devi

Because the com port description provided by the windows registry is not the usb description we often need, I refer to the above to get the device description of the usb serial port via DeviceIoControl.

#637

### DIFF
--- a/serial/tools/list_ports_windows.py
+++ b/serial/tools/list_ports_windows.py
@@ -21,8 +21,10 @@ from ctypes.wintypes import LONG
 from ctypes.wintypes import ULONG
 from ctypes.wintypes import HKEY
 from ctypes.wintypes import BYTE
+from ctypes.wintypes import PULONG
 import serial
 from serial.win32 import ULONG_PTR
+from serial.win32 import CreateFileW, GENERIC_WRITE, OPEN_EXISTING, DeviceIoControl, CloseHandle
 from serial.tools import list_ports_common
 
 
@@ -40,7 +42,6 @@ PTSTR = ctypes.c_wchar_p
 LPDWORD = PDWORD = ctypes.POINTER(DWORD)
 #~ LPBYTE = PBYTE = ctypes.POINTER(BYTE)
 LPBYTE = PBYTE = ctypes.c_void_p        # XXX avoids error about types
-
 ACCESS_MASK = DWORD
 REGSAM = ACCESS_MASK
 
@@ -73,6 +74,54 @@ class SP_DEVINFO_DATA(ctypes.Structure):
 
     def __str__(self):
         return "ClassGuid:{} DevInst:{}".format(self.ClassGuid, self.DevInst)
+
+
+class SetupPacket(ctypes.Structure):
+    _pack_ = 1
+    _fields_ = [
+        ('bmRequest', ctypes.c_uint8),
+        ('bRequest', ctypes.c_uint8),
+        ('wValue', ctypes.c_uint16),
+        ('wIndex', ctypes.c_uint16),
+        ('wLength', ctypes.c_uint16),
+    ]
+
+
+class USB_DESCRIPTOR_REQUEST(ctypes.Structure):
+    _pack_ = 1
+    _fields_ = [
+        ('ConnectionIndex', ctypes.c_uint32),
+        ('SetupPacket', SetupPacket),
+    ]
+
+
+class USB_DEVICE_DESCRIPTOR(ctypes.Structure):
+    _pack_ = 1
+    _fields_ = [
+        ('bLength', ctypes.c_uint8),
+        ('bDescriptorType', ctypes.c_uint8),
+        ('bcdUSB', ctypes.c_uint16),
+        ('bDeviceClass', ctypes.c_uint8),
+        ('bDeviceSubClass', ctypes.c_uint8),
+        ('bDeviceProtocol', ctypes.c_uint8),
+        ('bMaxPacketSize0', ctypes.c_uint8),
+        ('idVendor', ctypes.c_uint16),
+        ('idProduct', ctypes.c_uint16),
+        ('bcdDevice', ctypes.c_uint16),
+        ('iManufacturer', ctypes.c_uint8),
+        ('iProduct', ctypes.c_uint8),
+        ('iSerialNumber', ctypes.c_uint8),
+        ('bNumConfigurations', ctypes.c_uint8),
+    ]
+
+
+class USB_STRING_DESCRIPTOR(ctypes.Structure):
+    _pack_ = 1
+    _fields_ = [
+        ('bLength', ctypes.c_uint8),
+        ('bDescriptorType', ctypes.c_uint8),
+        ('bString', ctypes.c_wchar * 255),
+    ]
 
 
 PSP_DEVINFO_DATA = ctypes.POINTER(SP_DEVINFO_DATA)
@@ -131,6 +180,10 @@ CM_MapCrToWin32Err = cfgmgr32.CM_MapCrToWin32Err
 CM_MapCrToWin32Err.argtypes = [DWORD, DWORD]
 CM_MapCrToWin32Err.restype = DWORD
 
+CM_Get_DevNode_Status = cfgmgr32.CM_Get_DevNode_Status
+CM_Get_DevNode_Status.argtypes = [PULONG, PULONG, DWORD, ULONG]
+CM_Get_DevNode_Status.restype = DWORD
+
 
 DIGCF_PRESENT = 2
 DIGCF_DEVICEINTERFACE = 16
@@ -139,11 +192,16 @@ ERROR_INSUFFICIENT_BUFFER = 122
 ERROR_NOT_FOUND = 1168
 SPDRP_HARDWAREID = 1
 SPDRP_FRIENDLYNAME = 12
+SPDRP_ADDRESS = 28
 SPDRP_LOCATION_PATHS = 35
 SPDRP_MFG = 11
 DICS_FLAG_GLOBAL = 1
 DIREG_DEV = 0x00000001
 KEY_READ = 0x20019
+USB_REQUEST_GET_DESCRIPTOR = 6
+USB_DEVICE_DESCRIPTOR_TYPE = 1
+USB_STRING_DESCRIPTOR_TYPE = 3
+IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION = 2229264
 
 
 MAX_USB_DEVICE_TREE_TRAVERSAL_DEPTH = 5
@@ -233,6 +291,114 @@ def get_parent_serial_number(child_devinst, child_vid, child_pid, depth=0, last_
     # Finally, the VID and PID are identical to the child and a serial number is present, so return
     # it.
     return serial_number
+
+
+def request_usb_string_description(h_hub_device, usb_hub_port, idx):
+    if idx == 0:
+        return None
+    description_request_buffer = ctypes.create_string_buffer(ctypes.sizeof(USB_DESCRIPTOR_REQUEST) + ctypes.sizeof(USB_STRING_DESCRIPTOR))
+    description_request = ctypes.cast(description_request_buffer, ctypes.POINTER(USB_DESCRIPTOR_REQUEST))
+    description_request.contents.ConnectionIndex = usb_hub_port
+    description_request.contents.SetupPacket.wValue = (USB_STRING_DESCRIPTOR_TYPE << 8) | idx
+    description_request.contents.SetupPacket.wIndex = 0
+    description_request.contents.SetupPacket.wLength = ctypes.sizeof(description_request_buffer) - ctypes.sizeof(USB_DESCRIPTOR_REQUEST)
+    if not DeviceIoControl(
+            h_hub_device,
+            IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION,
+            description_request_buffer,
+            ctypes.sizeof(description_request_buffer),
+            description_request_buffer,
+            ctypes.sizeof(description_request_buffer),
+            None,
+            None
+    ):
+        return None
+    description = ctypes.cast(ctypes.byref(description_request_buffer, ctypes.sizeof(USB_DESCRIPTOR_REQUEST)),
+                              ctypes.POINTER(USB_STRING_DESCRIPTOR))
+    return ctypes.wstring_at(description.contents.bString, description.contents.bLength // 2 - 1)
+
+
+def request_usb_description(g_hdi, devinfo, info):
+    # Get usb port
+    usb_hub_port = DWORD()
+    if not SetupDiGetDeviceRegistryProperty(
+            g_hdi,
+            ctypes.byref(devinfo),
+            SPDRP_ADDRESS,
+            None,
+            ctypes.byref(usb_hub_port),
+            ctypes.sizeof(usb_hub_port),
+            None):
+        return False
+
+    # Get hub instance number
+    child_instance_number = devinfo.DevInst
+    while True:
+        node_status = ULONG()
+        problem_number = ULONG()
+        if CM_Get_DevNode_Status(ctypes.byref(node_status), ctypes.byref(problem_number), child_instance_number, 0) != 0:
+            return False
+        parent_instance_number = DWORD()
+        if CM_Get_Parent(ctypes.byref(parent_instance_number), child_instance_number, 0) != 0:
+            return False
+        parent_instance_id = ctypes.create_unicode_buffer(250)
+        if CM_Get_Device_IDW(
+            parent_instance_number,
+            parent_instance_id,
+            ctypes.sizeof(parent_instance_id) - 1,
+            0
+        ) != 0:
+            return False
+        parent_instance_id = ctypes.wstring_at(parent_instance_id)
+        if parent_instance_id.startswith('USB\\ROOT_HUB'):
+            break
+        else:
+            child_instance_number = parent_instance_number
+
+    # Generate the hub path and open it.
+    hub_location = "\\\\?\\" + parent_instance_id.replace("\\", "#") + "#{f18a0e88-c30c-11d0-8815-00a0c906bed8}"
+    h_hub_device = CreateFileW(
+        hub_location,
+        GENERIC_WRITE,
+        2,
+        None,
+        OPEN_EXISTING,
+        0,
+        0
+    )
+    if h_hub_device == INVALID_HANDLE_VALUE:
+        return False
+
+    # Send usb description request.
+    description_request_buffer = ctypes.create_string_buffer(ctypes.sizeof(USB_DESCRIPTOR_REQUEST) + ctypes.sizeof(USB_DEVICE_DESCRIPTOR))
+    description_request = ctypes.cast(description_request_buffer, ctypes.POINTER(USB_DESCRIPTOR_REQUEST))
+    description_request.contents.SetupPacket.bmRequest = 0x80
+    description_request.contents.SetupPacket.bRequest = USB_REQUEST_GET_DESCRIPTOR
+    description_request.contents.ConnectionIndex = usb_hub_port
+    description_request.contents.SetupPacket.wValue = USB_DEVICE_DESCRIPTOR_TYPE << 8
+    description_request.contents.SetupPacket.wLength = ctypes.sizeof(USB_DEVICE_DESCRIPTOR)
+    if not DeviceIoControl(
+        h_hub_device,
+        IOCTL_USB_GET_DESCRIPTOR_FROM_NODE_CONNECTION,
+        description_request_buffer,
+        ctypes.sizeof(description_request_buffer),
+        description_request_buffer,
+        ctypes.sizeof(description_request_buffer),
+        None,
+        None
+    ):
+        return False
+
+    # Get description
+    description = ctypes.cast(ctypes.byref(description_request_buffer, ctypes.sizeof(USB_DESCRIPTOR_REQUEST)),
+                              ctypes.POINTER(USB_DEVICE_DESCRIPTOR))
+    info.pid = description.contents.idProduct
+    info.vid = description.contents.idVendor
+    info.product = request_usb_string_description(h_hub_device, usb_hub_port, description.contents.iProduct)
+    info.manufacturer = request_usb_string_description(h_hub_device, usb_hub_port, description.contents.iManufacturer)
+    info.serial_number = request_usb_string_description(h_hub_device, usb_hub_port, description.contents.iSerialNumber)
+    CloseHandle(h_hub_device)
+    return True
 
 
 def iterate_comports():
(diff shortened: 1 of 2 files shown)


chinaheyu:
> I also tested it on Windows 10 and got the same result. I'm attaching my usbview report from Windows 11, in case you'd like to investigate, why I didn't get the expected result.
> [USBViewAll.txt](https://github.com/pyserial/pyserial/files/14089859/USBViewAll.txt)
> 

By the way, are you using the latest commit (`c090f45667c68b7a7259a38de7a59b49b016f851`)?

aivarannamaa:
> By the way, are you using the latest commit (c090f45667c68b7a7259a38de7a59b49b016f851)?

Yes, I was using https://github.com/pyserial/pyserial/blob/c090f45667c68b7a7259a38de7a59b49b016f851/serial/tools/list_ports_windows.py and https://github.com/pyserial/pyserial/blob/c090f45667c68b7a7259a38de7a59b49b016f851/serial/win32.py for patching my Pyserial.

I'll see later, whether the patched version works with my other devices.

chinaheyu:
> > By the way, are you using the latest commit (c090f45667c68b7a7259a38de7a59b49b016f851)?
> 
> Yes, I was using https://github.com/pyserial/pyserial/blob/c090f45667c68b7a7259a38de7a59b49b016f851/serial/tools/list_ports_windows.py and https://github.com/pyserial/pyserial/blob/c090f45667c68b7a7259a38de7a59b49b016f851/serial/win32.py for patching my Pyserial.
> 
> I'll see later, whether the patched version works with my other devices.

I found a similar issue here: https://stackoverflow.com/questions/51025878/how-to-get-string-descriptor-from-a-usb-device-which-is-in-low-power-state

Maybe it's because the device is in a low power state, so it can't communicate properly through IOCTL to get the usb descriptor (as you can see from the usbview output).

I'll try to reproduce and fix this issue later, thank you very much!

aivarannamaa:
I now tested your patch with more devices (SAMD-s from Adafruit, Rasperry Pi Pico, several ESP32-S2 and S3 devices, some plain ESP32 devices with CP and CH UART adapters, an ESP8266 with CP2104 UART adapter, a NRF52 device) and found out that with most of these it worked nicely.

The few, that showed proper Product (and Manufacturer) in usbview.exe, but not with your patch, were:

* Lilygo/TTGO T-Micro 32 ESP32 device (expected product value: "CP2104 USB to UART Bridge Controller", actual: None)
* Lolin C3 mini ESP32-C3 (expected product value: "USB JTAG/serial debug unit", actual: None). Curiously, your patch does enhance the value of the Manufacturer field, by changing it from "Microsoft" to "Espressif".
* M5Stack AtomU and AtomLite, ESP32 devices with FT232 or FT234 serial converter (expected value "M5Stack")
* Sparkfun ESP32 Thing (FT X-Series serial converter)
* BBC micro:bit v2. Besides not showing proper Product and Manufacturer fields, your patch turned serial_number field from "99063602000528203977597D73765584000000006E052820" to "" (empty string)

chinaheyu:
> I now tested your patch with more devices (SAMD-s from Adafruit, Rasperry Pi Pico, several ESP32-S2 and S3 devices, some plain ESP32 devices with CP and CH UART adapters, an ESP8266 with CP2104 UART adapter, a NRF52 device) and found out that with most of these it worked nicely.
> 
> The few, that showed proper Product (and Manufacturer) in usbview.exe, but not with your patch, were:
> 
> * Lilygo/TTGO T-Micro 32 ESP32 device (expected product value: "CP2104 USB to UART Bridge Controller", actual: None)
> * Lolin C3 mini ESP32-C3 (expected product value: "USB JTAG/serial debug unit", actual: None). Curiously, your patch does enhance the value of the Manufacturer field, by changing it from "Microsoft" to "Espressif".
> * M5Stack AtomU and AtomLite, ESP32 devices with FT232 or FT234 serial converter (expected value "M5Stack")
> * Sparkfun ESP32 Thing (FT X-Series serial converter)
> * BBC micro:bit v2. Besides not showing proper Product and Manufacturer fields, your patch turned serial_number field from "99063602000528203977597D73765584000000006E052820" to "" (empty string)
> 
> Please let me know, if you need usbview report for any of these!

I read the USBView report you provided for the BBC micro:bit v2. It looks like my patch misidentifies the composite device as a noncomposite device. This will result in an error similar to the one below.

Reads the string description as a composite device:

```
device: COM4
name: COM4
hwid: USB VID:PID=303A:4001 SER=123456 LOCATION=1-4:1.0
vid: 303a
pid: 4001
serial_number: 123456
location: 1-4:1.0
manufacturer: Espressif Systems
product: Espressif Device
```

Reads the string description as a noncomposite device:

```
device: COM4
name: COM4
hwid: USB VID:PID=303A:4001 SER=123456 LOCATION=1-4:x.0
vid: 303a
pid: 4001
serial_number: 
location: 1-4:x.0
manufacturer: Microsoft
```

Could you provide me with reports from the [USB Device Tree Viewer](https://www.uwe-sieber.de/usbtreeview_e.html) for these devices? It could provide me with more information to solve the problem. Thanks a lot.

aivarannamaa:
Here come the reports from USB Device Tree Viewer:

* [M5Stack Atom U.txt](https://github.com/pyserial/pyserial/files/14112481/M5Stack.Atom.U.txt)
* [M5Stack AtomLite.txt](https://github.com/pyserial/pyserial/files/14112482/M5Stack.AtomLite.txt)
* [Sparfun Thing 32.txt](https://github.com/pyserial/pyserial/files/14112483/Sparfun.Thing.32.txt)
* [BBC MicroBit.txt](https://github.com/pyserial/pyserial/files/14112484/BBC.MicroBit.txt)
* [Lilygo_TTGO T-Micro 32 ESP32.txt](https://github.com/pyserial/pyserial/files/14112485/Lilygo_TTGO.T-Micro.32.ESP32.txt)
* [Lolin C3 mini ESP32-C3.txt](https://github.com/pyserial/pyserial/files/14112486/Lolin.C3.mini.ESP32-C3.txt)

aivarannamaa:
Thanks a lot! After your last commit, pyserial reports proper Product, Manufacturer and Serial number for my BBC micro:bit. I'll test with my other boards tomorrow.

Great job!

aivarannamaa:
With all my boards I now get proper values for Manufacurer, Product and Serial number. For getting proper Interface, I'm combining your solution with #571.

Thank you!

chinaheyu:
> With all my boards I now get proper values for Manufacurer, Product and Serial number. For getting proper Interface, I'm combining your solution with #571.
> 
> Thank you!

I refactored the code and added comments to enhance readability. Thank you for your testing, I think this patch is ready to be merged now.

chinaheyu:
> With all my boards I now get proper values for Manufacurer, Product and Serial number. For getting proper Interface, I'm combining your solution with #571.
> 
> Thank you!

I noticed that you want to merge with #571. I've added the feature to read the interface (via `CM_Get_DevNode_Property` instead of `SetupDiGetDeviceProperty`) in the latest commit. So merging #571 may cause a conflict.

I put the old code for getting usb info into [`get_usb_info_from_device_property()`](https://github.com/chinaheyu/pyserial/blob/master/serial/tools/list_ports_windows.py#L680) as a fallback option. I suggest you put the change #571 at the end of the [`get_usb_info_from_device_property()`](https://github.com/chinaheyu/pyserial/blob/master/serial/tools/list_ports_windows.py#L680) function. This avoids conflicts caused by modifying the `iterate_comports()` function at the same time.

aivarannamaa:
> I've added the feature to read the interface

That's great! Somehow I missed it. I tested it and it works beautifully. I don't need #571 anymore.

chinaheyu:
It is now possible to get the correct interface field on most devices.

waterfallwhitebread:
Unfortunately does not seem to work well on devices with multiple serial interfaces, like FT4232H that has 4 UARTs.

Unpatched result:
```
device: COM12
name: COM12
description: USB Serial Port (COM12)
hwid: USB VID:PID=0403:6011 SER=6
vid: 1027
pid: 24593
serial_number: 6
location: None
manufacturer: FTDI
product: None
interface: None
```

Patched result:
```device: COM12
name: USB Serial Port (COM12)
description: Quad RS232-HS - Quad RS232-HS
hwid: USB VID:PID=0403:6011 LOCATION=1-1:1.0
vid: 1027
pid: 24593
serial_number: None
location: 1-1:1.0
manufacturer: FTDI
product: Quad RS232-HS
interface: Quad RS232-HS
```

Location is the same for all COM Ports, e.g.
```
device: COM11
name: USB Serial Port (COM11)
description: Quad RS232-HS - Quad RS232-HS
hwid: USB VID:PID=0403:6011 LOCATION=1-1:1.0
vid: 1027
pid: 24593
serial_number: None
location: 1-1:1.0
manufacturer: FTDI
product: Quad RS232-HS
interface: Quad RS232-HS
```

So there is no way to distinguish between the COM ports by location. For linux this seems to have been solved in PR https://github.com/pyserial/pyserial/pull/141

Here's an example of szHardwareID_str: FTDIBUS\VID_0403+PID_6011+6&1AD5555C&0&1&1\0000

chinaheyu:
> Unfortunately does not seem to work well on devices with multiple serial interfaces, like FT4232H that has 4 UARTs.
> 
> Unpatched result:
> 
> ```
> device: COM12
> name: COM12
> description: USB Serial Port (COM12)
> hwid: USB VID:PID=0403:6011 SER=6
> vid: 1027
> pid: 24593
> serial_number: 6
> location: None
> manufacturer: FTDI
> product: None
> interface: None
> ```
> 
> Patched result:
> 
> ```
> name: USB Serial Port (COM12)
> description: Quad RS232-HS - Quad RS232-HS
> hwid: USB VID:PID=0403:6011 LOCATION=1-1:1.0
> vid: 1027
> pid: 24593
> serial_number: None
> location: 1-1:1.0
> manufacturer: FTDI
> product: Quad RS232-HS
> interface: Quad RS232-HS
> ```
> 
> Location is the same for all COM Ports, e.g.
> 
> ```
> device: COM11
> name: USB Serial Port (COM11)
> description: Quad RS232-HS - Quad RS232-HS
> hwid: USB VID:PID=0403:6011 LOCATION=1-1:1.0
> vid: 1027
> pid: 24593
> serial_number: None
> location: 1-1:1.0
> manufacturer: FTDI
> product: Quad RS232-HS
> interface: Quad RS232-HS
> ```
> 
> So there is no way to distinguish between the COM ports by location. For linux this seems to have been solved in PR #141
> 
> Here's an example of szHardwareID_str: FTDIBUS\VID_0403+PID_6011+6&1AD5555C&0&1&1\0000

Since the FTDI driver hides the location information, we had to do some dirty work to get the interface number. I referenced [libusbp](https://github.com/pololu/libusbp/blob/master/src/windows/interface_windows.c#L222) to parse the hardware id to get the interface number. But apparently it doesn't work.

Could you provide me with reports from the [USB Device Tree Viewer](https://www.uwe-sieber.de/usbtreeview_e.html) for your devices?

waterfallwhitebread:
Hey @chinaheyu, thanks a lot for reply and the reference to libusbp. Sure, here the reports from USB Device Tree Viewer (just the first 2 UARTs, as the others are accordingly):
image
image
image
image

aivarannamaa:
After the last change most of my devices seem to work same as before, but I get following exception with 3 of my ESP32-C3 devices:

```
  File "C:\python_stuff\thonny\thonny\vendored_libs\serial\tools\list_ports_windows.py", line 1182, in comports
    return list(iterate_comports())
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\python_stuff\thonny\thonny\vendored_libs\serial\tools\list_ports_windows.py", line 1156, in iterate_comports
    usb_info = port_device.get_usb_info(device_registry)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\python_stuff\thonny\thonny\vendored_libs\serial\tools\list_ports_windows.py", line 698, in get_usb_info
    language_id = hub_io.suggest_language_id(usb_hub_port)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\python_stuff\thonny\thonny\vendored_libs\serial\tools\list_ports_windows.py", line 870, in suggest_language_id
    return available_languages[0]
           ~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

Here is the UsbTreeview report for one of them: [SeeedXiaoESP32C3.txt](https://github.com/pyserial/pyserial/files/14736239/SeeedXiaoESP32C3.txt)

The weird thing is that I get this exception only sometimes, usually the first time I call `comports()` after plugging in the device -- the next time I call `comports()` it usually works without problems (but not always).

aivarannamaa:
@chinaheyu, after latest changes I don't get these errors anymore, but I now discovered that with any of these ESP32-C3 devices plugged in, I now _occasionally_ get long delays before the ports info gets returned.

It looks like there are two methods, which may cause the delay:

* `USBHubDeviceIOControl.request_supported_languages`
* `USBHubDeviceIOControl.request_usb_string_description`

The first call of either method gets stuck in `DeviceIoControl` for 24-30 seconds. The next calls return quickly. For example, if I skip calling `DeviceIoControl` in `request_supported_languages`, the first `request_usb_string_description` becomes slow, etc.

The problem is present also in your older commits, e.g. https://github.com/aivarannamaa/pyserial/commit/bdab39f9f1989daa2fce57b68926806d00b6d0bc#diff-f9966444f1fafc17f351672a463bdcff7833ad6c14a68d9432329a458d1608b0

The next call to `comports()` is quick, unless in the meanwhile I've connected to or disconnected from a serial port in another process.

Unfortunately I can't create a simple script for reproducing this. The earlier I call comports in my app, the less likely the delay becomes.

Can you recommend some tweaks I could try to learn more about this problem?

aivarannamaa:
Adding to my last comment -- `DeviceIoControl` call in `request_usb_connection_info` never becomes slow, even though it gets called before `request_supported_languages` and `request_usb_string_description`.

chinaheyu:
@aivarannamaa Thank you very much for your feedback.

This is a very peculiar issue that I haven't encountered before. Overall, the problem should be related to the device (ESP32-32 C3) as it seems not to respond with its string description promptly. Subsequent requests are fast because DeviceIoControl caches the result.

After receiving your feedback, I purchased a Seeed Xiao ESP32 C3 board, which should arrive today. I will do my best to resolve this peculiar issue.

Additionally, I used the same API as in USBView, so did you encounter this issue when using USBView?

chinaheyu:
@waterfallwhitebread Thank you for providing the screenshot. The latest patch should be able to correctly retrieve the location field of the FT4232H chip.

chinaheyu:
@aivarannamaa You are right, if another process opens and closes the serial port, it blocks DeviceIOControl. This phenomenon only occurs on ESP32 C3 devices.

I also found an issue on StackOverflow that may be relevant: https://stackoverflow.com/questions/6749681/freeze-on-serialport-open-deviceiocontrol-getcommstate-with-usbser-sys

I can't fix this, but we can cache the usb info to avoid calling DeviceIOControl every time. The latest commit should be helpful for your problem.

aivarannamaa:
@chinaheyu, thanks! Curiously, now I can't reproduce this problem anymore.

PapaNoyel:
Any chance it gets merged into pyserial ?
This PR solved a problem I had on windows to differentiate a Nordic PPK2 from a Nordic Dongle as they appeared the same with vanilla pyserial.

Thanks for this chinaheyu !

jepler:
Hi! Is there anything that Adafruit can do to move inclusion of this patch forward? It would be very helpful to us.

keirf:
I have a report that this method doesn't work with Windows 7 (https://github.com/keirf/greaseweazle/issues/472). Is that unexpected?

chinaheyu:
It should theoretically support Windows 7 (though I haven't tested it), as Cfgmgr32 API has been available since Windows 2000. I plan to test it on Windows 7 next to see if I can reproduce the issue. Thank you @keirf .

chinaheyu:
> I have a report that this method doesn't work with Windows 7 ([keirf/greaseweazle#472](https://github.com/keirf/greaseweazle/issues/472)). Is that unexpected?

Screenshot from 2024-08-21 20-27-10

Sorry, I can't reproduce the problem. The patch works fine in Windows 7. Can you provide more detailed python version and system information?

The system image used above: Windows 7 Ultimate with Service Pack 1 (7601.24214) (x64)

chinaheyu:
Had some difficulty installing python 3.8, so used python 3.6, but the result should be the same.

Screenshot from 2024-08-21 21-25-21

32-bit Windows 7 and 32-bit Python

keirf:
Okay, I went to the hassle of installing Win7 in a VM to look at this, and what a pain that was! But I *have* reproduced the issue. I think it must depend on the device or its driver. Specifically, my "Greaseweazle" device enumerates as a generic CDC ACM device, and on Windows 7/8 requires use of Zadig to associate the driver usbser.sys (Zadig calls this "USB Serial (CDC)").

Curiously, the default pyserial list_port_windows.py finds the correct product string for this device on Windows 7 ("Greaseweazle"), which it fails to do on Windows 10/11. And on Win7 your list_port_windows.py fails to list it at all. Perhaps it gets filtered out somewhere?

keirf:
I also confirmed this issue on Windows 8.1. This is perhaps more related to Zadig, or pre-Win10 usbser.sys.

chinaheyu:
@keirf Bug maybe fixed. Could you try the latest commit  (91177bd)?

keirf:
> @keirf Bug maybe fixed. Could you try the latest commit ([91177bd](https://github.com/pyserial/pyserial/commit/91177bd485ab595aee83fb372b89fb2da477dee6))?

It works great! Thanks for the very quick fix. I have already ported it into Greaseweazle tools and I will have the original reporter also confirm the fix.

Thanks again!

adamcbraun:
Is there any update on this? This functionality (corrected device naming on windows) is actually addressed in a couple of the pending PRs, but does not seem to ever get completed. It would be great if one of these could reach completion as this functionality would be very helpful for a number of different of implementations.

keirf:
I have a report of failure with empty port name (https://github.com/keirf/greaseweazle/issues/549) with the following fix: https://github.com/keirf/greaseweazle/pull/550

Could you please consider this fix, or equivalent, for this upstream PR?

chinaheyu:
> I have a report of failure with empty port name ([keirf/greaseweazle#549](https://github.com/keirf/greaseweazle/issues/549)) with the following fix: [keirf/greaseweazle#550](https://github.com/keirf/greaseweazle/pull/550)
> 
> Could you please consider this fix, or equivalent, for this upstream PR?

It seems that some serial devices are not assigned a COMx-style port name. I'm not sure what's causing this, but simply skipping these devices can prevent the program from crashing. In theory, even without a port name, we can still open these serial devices via their interface path, which might serve as an alternative when the port name is None (possibly a better solution than just skipping them). That said, keirf/greaseweazle#550 is correct and necessary. Many thanks for the fix.

lambda-aw:
Hello,
Is there any progress on the issue?
The branch solves my issue with CP210x on Windows 11 and product field, so interested in any plan here.

Thanks.

EnzoVanhauwaert-Dekimo:
Hello,

This also solved my issue on Windows 11 and interface field, with a USB composite (2 CDC ACM) device.

Thank you.

alonbl:
Hi,

There is an issue in serial number.

Without patch:

```
VID:PID: E0D0:0002
Product: None
Serial: 0X829463B1-0X2A27980E
Interface: None
location: None
device: COM16
USB VID:PID=E0D0:0002 SER=0X829463B1-0X2A27980E
```

Linux
```
VID:PID: E0D0:0002
Product: mydevice
Serial: 0x829463b1-0x2a27980e
Interface: control (CIC)
location: 1-1:1.0
device: /dev/ttyACM0
USB VID:PID=E0D0:0002 SER=0x829463b1-0x2a27980e LOCATION=1-1:1.0
```

With patch:
```
VID:PID: E0D0:0002
Product: mydevice
Serial: 0x829463b1-0x2a27980e杩oꮐ翸臠咋Ơ䒀孒翹Ĝ孒翹齰翸￿￿￿￿
Interface: control (CIC)
location: 2-6.2.3:1.0
device: COM16
USB VID:PID=E0D0:0002 SER=0x829463b1-0x2a27980e杩oꮐ翸臠咋Ơ䒀孒翹Ĝ孒翹齰翸￿￿￿￿ LOCATION=2-6.2.3:1.0
```

Notice the serial number junk.

Thanks,
Alon

lukehugh:
Hey, @alonbl,

During decoding, we did not use decoding APIs like `WideCharToMultiByte`. We assumed the device returned the correct length and then converted it to a Python string using `ctypes.wstring_at`. The code looks something like this.

```python
# Convert wstring to python str.
# Very few devices will have an extra null at the end of string, strip it.
return ctypes.wstring_at(description.contents.bString, description.contents.bLength // 2 - 1).rstrip('\0')
```

Therefore, we are curious whether the device returned an incorrect bLength.

Could you download [usbtreeview](https://www.uwe-sieber.de/usbtreeview_e.html) and show me the serial number in the string description section? It should look something like this:

```
             ------ String Descriptor 3 ------
bLength                  : 0x24 (36 bytes)
bDescriptorType          : 0x03 (String Descriptor)
Language 0x0409          : "20120501030900000"
Data (HexDump)           : 24 03 32 00 30 00 31 00 32 00 30 00 35 00 30 00   $.2.0.1.2.0.5.0.
                           31 00 30 00 33 00 30 00 39 00 30 00 30 00 30 00   1.0.3.0.9.0.0.0.
                           30 00 30 00                                       0.0.
```

Thanks.

alonbl:
Hi @lukehugh,

This is interesting!

I get the following:

```
             ------ String Descriptor 2 ------
bLength                  : 0x1A (26 bytes)
bDescriptorType          : 0x03 (String Descriptor)
Language 0x0409          : "xxxxxxxxxxxx"
             ------ String Descriptor 3 ------
bLength                  : 0x82 (130 bytes)
bDescriptorType          : 0x03 (String Descriptor)
Language 0x0409          : "829463b1-2a27980e°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°"  *!*ERROR  contains 47 NULL characters  *!*CAUTION  Windows uses only up to 63 characters from iSerial for building Device IDs of DISK devices
```

Update: I found the hex dump... and now I can see that the host indeed does not receive correct size, I will investigate, thank you for your help!

```
             ------ String Descriptor 3 ------
bLength                  : 0x82 (130 bytes)
bDescriptorType          : 0x03 (String Descriptor)
Language 0x0409          : "829463b1-2a27980e°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°"  *!*ERROR  contains 47 NULL characters  *!*CAUTION  Windows uses only up to 63 characters from iSerial for building Device IDs of DISK devices
Data (HexDump)           : 82 03 38 00 32 00 39 00 34 00 36 00 33 00 62 00   ..8.2.9.4.6.3.b.
                           31 00 2D 00 32 00 61 00 32 00 37 00 39 00 38 00   1.-.2.a.2.7.9.8.
                           30 00 65 00 00 00 00 00 00 00 00 00 00 00 00 00   0.e.............
                           00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
                           00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
                           00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
                           00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
                           00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
                           00 00                                             ..
```

Thanks,
Alon

alonbl:
Hi @lukehugh ,

When using Linux and libusb, I get:

```
iSerialNumber index: 3
Raw serial string descriptor bytes:
24 03 38 00 32 00 39 00 34 00 36 00 33 00 62 00 31 00 2d 00 32 00 61 00 32 00 37 00 39 00 38 00 30 00 65 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
```

I do not understand why in windows we get 0x82 which is the buffer size and in Linux we get 0x24 which is the string size. Maybe there usbview is not showing the correct values?

Thanks,

lukehugh:
@alonbl Although I can't do anything about what your device returns on Windows. Theoretically, if the string contains extra NULL characters, we should strip them (and indeed, the code does this). Perhaps your Windows PC uses some special code page that doesn't interpret NULL as '\0', so I explicitly used UTF-16LE encoding in the latest commit (00961acca52e8270eda768a1a97078ff0c55e369). This might help you.

alonbl:
Hi @lukehugh,
I found that windows overwrites the size byte as if it was sent from device.
In both Linux and MacOS this is not happening.
I modified the device to adjust the capacity as well, so it is working with this implementation as well.
There should be no extra NULL in USB descriptor strings as far as I know.
I checked the last commit and I can confirm it is working.
What are the chances this get merge? It has been two years since you submitted.
Thanks,
Alon

lambda-aw:
hello,
there will be joy on my side once this is eventually merged and released (when?), so the `serial` returns finally something sensible.
if you go to wait for new year, I'll be back at work and can quickly test it with our internal USB device, if it behaves nicelly too on Windows and Debian at the same time.
thanks.

Jaroslav

lukehugh:
Regarding whether this PR will be merged, I’d like to give a unified response here.

This patch originated from one of my personal side projects. At the time, I was struggling to reliably obtain the correct *product string* (since registering a USB VID is expensive, we wanted to differentiate devices via the product string instead). Later, I discovered **USBView** and learned how to retrieve proper USB descriptors using `DeviceIoControl`. To achieve a correct product string, we replaced the outdated `setupapi` with `cfgmgr32`, and added a thin wrapper around the Device Node and `DeviceIoControl`. That resulted in my first working version. While the implementation was admittedly minimal, it was sufficient to solve my problem.

Considering that other users in the pyserial community might have similar needs—and hoping this approach could benefit from broader validation—we decided to submit this PR. With the help of several enthusiastic contributors, it gradually evolved. That said, I am not a pyserial maintainer, so I am not in a position to decide whether this PR should ultimately be merged. Unfortunately, around that time pyserial also appeared to be unmaintained. As a result, I chose to make the patch as self-contained as possible, so users could simply copy the file into their own projects as a workaround. In fact, the code in `list_ports_windows.py` only depends on `list_ports_common.py`, and even that dependency could be removed if necessary.

Today, I’m happy to see that pyserial is active again and gradually receiving new commits. However, I remain uncertain about the future of this PR. From my perspective, the original problem has already been solved; for the community, integrating this patch into other projects is relatively straightforward; but for pyserial itself, merging this PR could introduce breaking changes. In any case, I would very much appreciate hearing from the maintainers. @dpgeorge, @kurtmckee.

Thanks everyone for the attention and discussion.

kurtmckee:
Hi @lukehugh, thanks for the ping. I'm also not a pyserial maintainer.

I've been holding off on reviewing this PR for only one reason: I'm waiting for #838 to merge. #838 introduces a command to generate a "test asset", which is a JSON file that aims to contain all of the available information for a given device (or devices, if needed) on a given system. That's useful both for debugging detection issues (like what @alonbl reported with trailing null characters) as well as for mocking system calls to confirm that pyserial is handling the devices properly.

#838 introduces such files for sysfs-based operating systems (like Linux), but I've intended to expand the command to support Windows once the PR merges, so that I could ask @alonbl and yourself to please generate the test assets for the devices you're working with. With those test asset files in-hand, as well as files from my own devices on Windows, I've intended to build out the Windows side of the test suite.

The Windows code in pyserial is completely untested, and your PR is a significant contribution that inherits that complete lack of testing. I've therefore been waiting to dive into a review of the PR until I can build out some level of device mocking and testing to better understand the before-and-after behavior of your PR.

I hope this explanation is helpful for understanding why I haven't reviewed the PR yet.

alonbl:
Hello @kurtmckee ,

There is always a question of CI and hardware tests....

Do we gain value from simulating hardware for synthetic tests or we invest large amount of resources and eventually we do not test the real-world scenarios.

In case of sysfs, it is quite simple as the sysfs abstraction is plain files, so you could commit the relevant sysfs as-is, and only have a `SYSFS_ROOT` environment variable (for example) to scan the custom directory even without monkey patching the code... So you can have one or two tests for sysfs to make sure we read it correctly... this is not all-on-one-multi-platform agenda, just a simple unit test for sysfs.

In case of windows, it is more difficult, as Windows APIs are much more complex. However, these are static, and decent manual tests of the low-level parts should be sufficient.

For sure, having CI or not should not block us from reviewing code...

Regards,
Alon

kurtmckee:
> For sure, having CI or not should not block us from reviewing code...

You are correct. I look forward to seeing your review of the code in this PR!

PR #732, opened 2024-01-22
Set  RTS and DTR lines at the same time when they are the same value 

### LINKED ISSUE #731: DTR and RST set at different times
When setting the DTR and RST values to the same value, they output signals do not get set at the same time. This causes problem in program that use them to reset and flash micr controllers.

fluffy:
Fixes #731 

When both the DTR and RST lines are bien get to the same value, this PR will cause them to be set at the same time instead of setting one of them followed by the other.

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -283,6 +283,7 @@ TIOCMSET = getattr(termios, 'TIOCMSET', 0x5418)
 # TIOCM_LE = getattr(termios, 'TIOCM_LE', 0x001)
 TIOCM_DTR = getattr(termios, 'TIOCM_DTR', 0x002)
 TIOCM_RTS = getattr(termios, 'TIOCM_RTS', 0x004)
+TIOCM_DTRRTS = getattr(termios, 'TIOCM_DTR', 0x002) | getattr(termios, 'TIOCM_RTS', 0x004)
 # TIOCM_ST = getattr(termios, 'TIOCM_ST', 0x008)
 # TIOCM_SR = getattr(termios, 'TIOCM_SR', 0x010)
 
@@ -303,6 +304,7 @@ TIOCOUTQ = getattr(termios, 'TIOCOUTQ', 0x5411)
 TIOCM_zero_str = struct.pack('I', 0)
 TIOCM_RTS_str = struct.pack('I', TIOCM_RTS)
 TIOCM_DTR_str = struct.pack('I', TIOCM_DTR)
+TIOCM_DTRRTS_str = struct.pack('I', TIOCM_DTRRTS)
 
 TIOCSBRK = getattr(termios, 'TIOCSBRK', 0x5427)
 TIOCCBRK = getattr(termios, 'TIOCCBRK', 0x5428)
@@ -339,10 +341,13 @@ class Serial(SerialBase, PlatformSpecific):
             self._reconfigure_port(force_update=True)
 
             try:
-                if not self._rtscts:
-                    self._update_rts_state()
-                if not self._dsrdtr:
-                    self._update_dtr_state()
+                if (not self._rtscts) and (not self._dsrdtr):
+                    self._update_rts_and_dtr_state()
+                else:
+                    if not self._rtscts:
+                        self._update_rts_state()
+                    if not self._dsrdtr:
+                        self._update_dtr_state()
             except IOError as e:
                 # ignore Invalid argument and Inappropriate ioctl
                 if e.errno not in (errno.EINVAL, errno.ENOTTY):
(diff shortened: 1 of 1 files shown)


fluffy:
any chance someone could take over as maintainer ?

serial/serialposix.py
```diff
@@ -707,6 +712,21 @@ def send_break(self, duration=0.25):
             raise PortNotOpenError()
         termios.tcsendbreak(self.fd, int(duration / 0.25))
 
+    def _update_rts_and_dtr_state(self):
```
kurtmckee: It looks like `._update_rts_and_dtr_state()` can supersede the functionality of both `._update_rts_state()` and `._update_dtr_state()`. Is that an accurate assessment?

If so, please eliminate both of the individual update methods in favor of this one. That may reduce the complexity of the code in `.open()`:

```python
if (not self._rtscts) or (not self._dsrdtr):
    self._update_rts_and_dtr_state()
```

fluffy:
@kurtmckee - the issue is no one is merging any of the PR into mainline so not much point in fixing this up

kurtmckee:
30 PRs have merged or closed in the last two weeks, and ~30 issues have been closed or fixed as well.

That doesn't mean that the suggestion above should be taken, nor that this PR will be merged by the maintainers. However, I'm working to drive issues and PRs to closure, and this one is easy to review. Plus, you're responsive! :partying_face:

dpgeorge:
I support this change.

serial/serialposix.py
```diff
@@ -345,10 +348,7 @@ def open(self):
             self._reconfigure_port(force_update=True)
 
             try:
-                if not self._dsrdtr:
-                    self._update_dtr_state()
-                if not self._rtscts:
-                    self._update_rts_state()
+                self._update_rts_and_dtr_state()
```
fluffy: I think I need to look at this a bit more. The change here would allow the rts and dtr io lines to be set even when in one of the flow control modes. That is probably wrong.

fluffy:
@kurtmckee - Very much appreciated if you get to have a look at this. As I went to test, it expanded into adding an public api to set the rts/dsr at same time.

serial/serialposix.py
```diff
@@ -748,6 +748,27 @@ def send_break(self, duration=0.25):
...
+                fcntl.ioctl(self.fd, TIOCMSET, struct.pack('I', v))
+        else:
+            if self._rts_state:
+                s = fcntl.ioctl(self.fd, TIOCMGET, TIOCM_all_str)
+                v = struct.unpack('I', s)[0]
+                v &= ~TIOCM_DTR
+                v |= TIOCM_RTS
+                fcntl.ioctl(self.fd, TIOCMSET, struct.pack('I', v))
+            else:
+                fcntl.ioctl(self.fd, TIOCMBIC, TIOCM_DTRRTS_str)
```
kurtmckee: I think this formulation is more readable, but please verify that it's actually equivalent! What are your thoughts?


```suggestion
        s = fcntl.ioctl(self.fd, TIOCMGET, TIOCM_all_str)
        v = struct.unpack('I', s)[0]

        # Set or clear DTR.
        if self._dtr_state:
            v |= TIOCM_DTR
        else:
            v &= ~TIOCM_DTR

        # Set or clear RTS.
        if self._rts_state:
            v |= TIOCM_RTS
        else:
            v &= ~TIOCM_RTS

        fcntl.ioctl(self.fd, TIOCMSET, struct.pack('I', v))
```

serial/serialutil.py
```diff
@@ -476,6 +476,21 @@ def dtr(self, value):
...
+    def _update_rts_and_dtr_state(self):
+        self._update_rts_state()
+        self._update_dtr_state()
+
+    @property
+    def rtsdtr(self):
+        return (self._rts_state,self._dtr_state)
+
+    @rtsdtr.setter
+    def rtsdtr(self, value):
```
kurtmckee: I think that following the property-based setting method that exists for configuring individual values establishes an incorrect contract for developers to work with. Instead, this needs to be a function that allows _all_ values (except port, for example, which would fundamentally alter the `SerialBase` instance) to be configured simultaneously, and _using keyword parameters exclusively_. The initial implementation could simply support `rts` and `dts`, and be expanded in subsequent work.

For example:

```python
def reconfigure_port(
    self,
    *,
    rts: bool | None = None,
    dtr: bool | None = None,
) -> None:
    if rts is not None:
        self._rts_state = rts
    if dts is not None:
        self._dts_state = dts

    if self.is_open:
        self._reconfigure_port()
```

@zsquareplusc The goal of this PR is to support setting RTS and DTR at the same time, both as an interface that developers can use, and at a system call level for platforms that support it.

Would you take a look at my suggestion in the code example above, which addresses the interface that developers use?

PR #643, opened 2022-03-27
Fix threading module related DeprecationWarning.

tirkarthi:
The camelcase methods were deprecated in Python 3.10 in https://github.com/python/cpython/pull/25174

PR #734, opened 2024-01-30
treewide: fix remaining uses of threading camel case methods

Bastian-Krause:
The threading camel case methods were deprecated in https://github.com/python/cpython/pull/25174. PR #643 fixed a bunch of these warnings, but missed a few.

Fix the remaining `setName()` DeprecationWarnings by using the appropriate `name` attribute instead.


Bastian-Krause:
@zsquareplusc @bufferoverflow I think there's not much to argue over. Could you please merge this?

PR #747, opened 2024-03-19
Update shortintro.rst

alanmehio:
Suggesting changes to work on python3 
The existing code works on python2 but not python3 . Please notice

The character encoding can be defined in the TextIOWrapper instead otherwise if not given in TextIOWrapper it will be None 
The code sample can be split into python2 and python3 as  an additional note.


zsquareplusc:
Updating for Py 3 syntax is actually good :-) thanks

PR #758, opened 2024-06-10
Fix for Interface always None on Windows

ben-qnimble:
When listing comports on Windows, the interface is always set to None. This is because in list_ports_windows.py the bInterfaceNumber is correctly determined and set locally with the variable bInterfaceNumber.  But it is never set to the object, so when serial.tools.list_ports.comports() is run, interface is always None. 

This one-line addition fixes that.

### DIFF
--- a/serial/tools/list_ports_windows.py
+++ b/serial/tools/list_ports_windows.py
@@ -364,6 +364,7 @@ def iterate_comports():
                                 location.append('-')
                             location.append(g.group(2))
                     if bInterfaceNumber is not None:
+                        info.interface = bInterfaceNumber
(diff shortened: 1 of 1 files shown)


alonbl:
This is a great fix, indeed missing.

However, it means that in Windows we get a numeric interface while in other platforms we get the string representation that is determined by the device.

If this the best we can do in Windows, that's ok... I suggest to have the string as `Unknown <n>` or something similar to allow distinguish that this was not obtained from the device.

alonbl:
Hi,
I found that #725 is a superset and also applies correct name similar to Linux.
Thanks,

PR #764, opened 2024-07-24
rfc2217: do not set logger = None inside open()

cdleonard:
The logger field is already initialized inside __init__ so no need to do it again.

It is possible for the user to initialize logger after __init__ and before open(), if that happens then we should log all negotiation messages. Right now they are always lost.


kurtmckee (approved):
@zsquareplusc @dpgeorge Please merge this PR.

It allows users to configure logging between instantiation and calling `.open()`, and allows a conditional logging line later in `.open()` (around line 483) to potentially execute.

Issue #343, opened 2018-05-18
KeyboardInterrupt on Windows during read


cmcqueen:
On Windows, if a serial port is set up for no timeout, and a `s.read()` is done with no data arriving, then it is not possible to interrupt the read using Ctrl-C. It would be good if the read would be interruptible by Ctrl-C, to generate a `KeyboardInterrupt`.

fdila:
I'm having the same problem on linux

betasilabs:
same here

JaewonJeung:
I had the same problem, but it seems like Ctrl + Pause/Break key works on Windows 10 Powershell.

jonathangjertsen:
My workaround for this issue was to make some wrapper functions for the read, read_until, readline and readlines methods in which the serial timeout is temporarily set to some short value (the temporary timeout should be short enough to get a fast response to the keyboard interrupts, and I guess it should be much longer than it takes to read a byte).

```Python
import time
import serial

TEMPORARY_TIMEOUT = 0.1

def interruptible_wrapper(ser: serial.Serial, func, timeout) -> bytes:
    original_timeout = ser.timeout
    data = b''
    try:
        ser.timeout = TEMPORARY_TIMEOUT
        start_time = time.time()
        while not data:
            data = func()
            if timeout is not None and time.time() - start_time > timeout:
                break
    except KeyboardInterrupt:
        raise KeyboardInterrupt("Interrupted during read") from None
    finally:
        ser.timeout = original_timeout
    return data


def read(ser: serial.Serial, size: int=1, timeout=None):
    return interruptible_wrapper(ser, lambda: ser.read(size=size), timeout=timeout)


def read_until(ser: serial.Serial, expected: bytes=serial.LF, size: int=None, timeout=None):
    # Deal with backwards incompatible change introduced in 3.5
    if serial.__version__ >= "3.5":
        kwargs = { "size": size, "expected": expected }
    else:
        kwargs = { "size": size, "terminator": expected }
    return interruptible_wrapper(ser, lambda: ser.read_until(**kwargs), timeout)


def readline(ser: serial.Serial, size: int=-1, timeout=None):
    return interruptible_wrapper(ser, lambda: ser.readline(size), timeout)


def readlines(ser: serial.Serial, hint: int=-1, timeout=None):
    return interruptible_wrapper(ser, lambda: ser.readlines(hint), timeout)
```

Usage:

```Python
print("read:", read(ser))
[...]

PR #770, opened 2024-09-04
Don't block KeyboardInterrupt with kernel mode indefinite wait on Win32

nkinnan:
If no data is coming in, `GetOverlappedResult` will wait in kernel mode indefinitely.  This prevents things like KeyboardInterrupt due to ctrl-c from happening.  

This change calls `GetOverlappedResultEx` instead which has a timeout parameter, and just re-waits every 100 milliseconds, returning from kernel mode in-between to allow exceptions to be processed.

This PR resolves issue https://github.com/pyserial/pyserial/issues/343 at least on windows but there was a complaint about the same issue on Linux which will need a similar fix.

### DIFF
--- a/serial/serialwin32.py
+++ b/serial/serialwin32.py
@@ -285,10 +285,12 @@ class Serial(SerialBase):
                     ctypes.byref(self._overlapped_read))
                 if not read_ok and win32.GetLastError() not in (win32.ERROR_SUCCESS, win32.ERROR_IO_PENDING):
                     raise SerialException("ReadFile failed ({!r})".format(ctypes.WinError()))
-                timeout = Timeout(self._timeout)
+                # some dependencies rely on the broken timeout behavior, so avoiding fixing that for now
+                # timeout = Timeout(self._timeout)
                 while True:
                     # even with timeout infinite, return from kernel mode every 100ms to allow for things like KeyboardInterrupt to happen
-                    millis = win32.DWORD(int(min(timeout.time_left() or 0.1, 0.1) * 1000))
+                    # millis = win32.DWORD(int(min(timeout.time_left() or 0.1, 0.1) * 1000))
+                    millis = 100
                     result_ok = win32.GetOverlappedResultEx(
                         self._port_handle,
                         ctypes.byref(self._overlapped_read),
@@ -297,8 +299,8 @@ class Serial(SerialBase):
                         False)
                     if result_ok or win32.GetLastError() != win32.WAIT_TIMEOUT:
                         break
-                    if timeout.expired():
-                        break
+                    # if timeout.expired():
+                    #     break
                 if not result_ok:
                     if win32.GetLastError() != win32.ERROR_OPERATION_ABORTED:
                         raise SerialException("GetOverlappedResult failed ({!r})".format(ctypes.WinError()))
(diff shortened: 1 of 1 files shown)


nkinnan:
Fixes issue https://github.com/pyserial/pyserial/issues/343

serial/serialwin32.py
```diff
@@ -316,9 +325,22 @@ def write(self, data):
                 if not success and win32.GetLastError() not in (win32.ERROR_SUCCESS, win32.ERROR_IO_PENDING):
                     raise SerialException("WriteFile failed ({!r})".format(ctypes.WinError()))
 
-                # Wait for the write to complete.
-                #~ win32.WaitForSingleObject(self._overlapped_write.hEvent, win32.INFINITE)
-                win32.GetOverlappedResult(self._port_handle, self._overlapped_write, ctypes.byref(n), True)
+                timeout = Timeout(self._timeout)
+                while True:
+                    # even with timeout infinite, return from kernel mode every 100ms to allow for things like KeyboardInterrupt to happen
+                    millis = win32.DWORD(int(min(timeout.time_left() or 0.1, 0.1) * 1000))
```
nkinnan, line 331: if time_left() is 0, millis will be 0
if time_left() is < 0.1, millis will be time_left() * 1000
if time_left() is > 0.1, millis will be 0.1 * 1000 (capped)

serial/serialwin32.py
```diff
@@ -285,11 +285,20 @@ def read(self, size=1):
...
+                    result_ok = win32.GetOverlappedResultEx(
+                        self._port_handle,
+                        ctypes.byref(self._overlapped_read),
+                        ctypes.byref(rc),
+                        millis,
+                        False)
+                    if result_ok or win32.GetLastError() != win32.WAIT_TIMEOUT:
```
nkinnan, line 298: non-zero result is success, in case of timeout we expect failure and a GetLastError value of WAIT_TIMEOUT

serial/serialwin32.py
```diff
@@ -285,11 +285,20 @@ def read(self, size=1):
...
                     raise SerialException("ReadFile failed ({!r})".format(ctypes.WinError()))
-                result_ok = win32.GetOverlappedResult(
-                    self._port_handle,
-                    ctypes.byref(self._overlapped_read),
-                    ctypes.byref(rc),
-                    True)
+                timeout = Timeout(self._timeout)
+                while True:
+                    # even with timeout infinite, return from kernel mode every 100ms to allow for things like KeyboardInterrupt to happen
+                    millis = win32.DWORD(int(min(timeout.time_left() or 0.1, 0.1) * 1000))
```
nkinnan, line 291: if time_left() is 0, millis will be 0
if time_left() is < 0.1, millis will be time_left() * 1000
if time_left() is > 0.1, millis will be 0.1 * 1000 (capped)

serial/serialwin32.py
```diff
@@ -285,11 +285,22 @@ def read(self, size=1):
                     ctypes.byref(self._overlapped_read))
                 if not read_ok and win32.GetLastError() not in (win32.ERROR_SUCCESS, win32.ERROR_IO_PENDING):
                     raise SerialException("ReadFile failed ({!r})".format(ctypes.WinError()))
-                result_ok = win32.GetOverlappedResult(
-                    self._port_handle,
-                    ctypes.byref(self._overlapped_read),
-                    ctypes.byref(rc),
-                    True)
+                # some dependencies rely on the broken timeout behavior, so avoiding fixing that for now
```
nkinnan, line 288: Current timeout behavior effectively pads the value, so fixing it to be accurate here actually broke some downstream dependencies.  So, unfortunately, while this still fixes the KeyboardInterrupt problem, I had to revert the timeout fix for compatibility reasons.

PR #778, opened 2024-10-22
FreeBSD implementation of list_ports

bsdphk:
This PR implements ``list_ports`` on FreeBSD using the ``devinfo(8)`` program.

The USB strings are unfortunately not available through ``devinfo(8)``.


sbz:
@bsdphk While waiting for a merge and a next release, I have included your PR in the port tree  https://github.com/freebsd/freebsd-ports/commit/d1a2a3d66e087b3bb91abb22a25f3d1ef4b4f4c7

bsdphk:
May I kindly ask that you merge this PR ?

sbz:
@dpgeorge @zsquareplusc Would you mind to help with @bsdphk request? It would be great to release a new version with the feature included

serial/tools/list_ports_freebsd.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python
+#
+# This is a module that gathers a list of serial ports including details on
+# GNU/Linux systems.
```
projectgus, line 4:
```suggestion
# FreeBSD systems.
```

serial/tools/list_ports_freebsd.py
```diff
@@ -0,0 +1,69 @@
...
+            self.apply_usb_info()
+        else:
+            self.subsystem = "uart"
+            self.hwid = self.description
+
+    def usb_description(self):
+        return self.props["ugen"]
+
+def comports(include_links=False):
+    x = subprocess.run(["/usr/sbin/devinfo", "-rv"], capture_output=True)
```
projectgus, line 52: Is the output of this tool considered stable?

It looks like the alternatives would to be pass `--libxo` to get JSON or XML output, or to bind and call the `devinfo` API directly?

I don't personally know a lot about FreeBSD, so I'm happy to do whatever makes the most sense in that context. (EDIT: including keeping this as-is.)

serial/tools/list_ports_posix.py
```diff
@@ -50,11 +50,7 @@ def comports(include_links=False):
         return [list_ports_common.ListPortInfo(d) for d in devices]
 
 elif plat[:3] == 'bsd' or plat[:7] == 'freebsd':
```
projectgus, line 52: Apologies for my lack of deep BSD lore here, but does anyone know the newest Python(s) and platform(s) that might match `bsd` in this case? [For CPython](https://docs.python.org/3/library/sys.html#sys.platform) (or [2.7](https://docs.python.org/2/library/sys.html#sys.platform)) this would be a platform for which the output of `uname -s` starts with "bsd" (case insensitive.) It's not OpenBSD, NetBSD or FreeBSD, obviously.

If such a platform exists and may run current pyserial, will it have a `/usr/sbin/devinfo` binary available?

If the answers are "this exists" and "no", respectively, then this probably needs to be split into two branches and keep the old code around. But if this is only for original original BSD, then maybe we can totally drop it.

projectgus (commented):
Thanks for being patient while someone got around to reviewing this.

I'm not a FreeBSD expert but given this is already in the FreeBSD ports tree it seems fine to me! I have a couple of minor questions, and want to check one detail about whether this might change behaviour on some other unknown BSD variant(s).

serial/tools/list_ports_freebsd.py
sbz: Thanks for the comments! Yes, indeed it's possible to retrieve JSON output on stdout when calling `devinfo -rv --libxo=json` but I don't think it's worth using it, the parsing is already done with plain text with [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) style fashion 

> _Expect the output of every program to become the input to another, as yet unknown, program. Don't clutter output with extraneous information. Avoid stringently columnar or binary input formats. Don't insist on interactive input.._ 

That's said, I don't think it will be too difficult to change it and parse the JSON output but unnecessary cc @bsdphk

serial/tools/list_ports_posix.py
sbz: [devinfo(8)](https://man.freebsd.org/cgi/man.cgi?devinfo) is FreeBSD specific here. Others BSDs such NetBSD or OpenBSD have something else to retrieve this information, so we could drop `plat[:3] == 'bsd'` here only `plat[:7] == 'freebsd'` is relevant IMHO.

```python
root@freebsd15:~ # python3.11 
Python 3.11.14 (main, Jan  4 2026, 01:06:39) [Clang 19.1.7 (https://github.com/llvm/llvm-project.git llvmorg-19.1.7-0-gcd7080 on freebsd15
>>> import sys
>>> sys.platform.lower()[:7]
'freebsd'
```

sbz:
Any chance to see this feature included in a new release soon? cc @projectgus

kev009:
@bsdphk @sbz would be good to get your eyes on the above, this PR doesn't work for me with greaseweazle

sbz:
@kev009 Looks good to me. Thanks for the updates! I think the strategy to parse the pnpinfo then fallback is more robust :muscle: and will support more devices. 

Hopefully we could have a merge and a new release soon  :pray:  cc @projectgus

PR #786, opened 2025-02-06
posix: Catch errno from termios.error

knuxify:
termios.error exceptions have an errno passed to the arguments.

Fetch the errno from the args and forward it to the SerialException.

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -405,7 +405,7 @@ class Serial(SerialBase, PlatformSpecific):
             orig_attr = termios.tcgetattr(self.fd)
             iflag, oflag, cflag, lflag, ispeed, ospeed, cc = orig_attr
         except termios.error as msg:      # if a port is nonexistent but has a /dev file, it'll fail here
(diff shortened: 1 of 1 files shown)


kurtmckee (approved):
@zsquareplusc @projectgus Please merge this PR. It improves how `SerialException` is instantiated and represented.

`SerialException` inherits from `OSError` and this change switches to [the second `OSError` instantiation signature](https://docs.python.org/3/library/exceptions.html#OSError):

```python
OSError(errno, strerror)
```

The result of this PR is that `SerialException` will now have a `.errno` attribute with a valid value, and a clearer string representation:

```pycon
>>> error = SerialException(9, "Could not configure port: (9, 'Bad file descriptor')")
>>> print(error.errno)
9
>>> print(error.strerror)
Could not configure port: (9, 'Bad file descriptor')
>>> print(error)
[Errno 9] Could not configure port: (9, 'Bad file descriptor')
```

Although this could be improved further by removing the duplication of the `errno`, that can happen in subsequent work, and this PR can be merged as a standalone improvement.

Issue #396, opened 2018-12-09
inter_byte_timeout deserves better treatment


rdpoor:
## Statement of the problem:

Modbus uses an inter-packet delay of 3.5 byte periods to detect the end of a packet.  One might reasonably expect that `inter_byte_timeout` is suited for this task, along the lines of:

    n_bits = 8 + stopbits
    ibt_s = n_bits*3.5/baudrate
    ser = Serial(port=port, baudrate=baudrate, stopbits=stopbits, inter_byte_timeout=ibt_s, ...)

## But...

There are a few problems with the above:
* Windows expects `inter_byte_timeout` to be expressed in milliseconds
* POSIX expects `inter_byte_timeout` to be expressed in tenths of seconds
* Java and .NET backends don't support `inter_byte_timeout` at all
* ... and none of these nuances are documented

## Recommendations

Judging from the comments in this repository and StackOverflow, better documentation of `inter_byte_timeout` would prevent confusion and frustration.  And a better implementation would make pySerial more useful for Modbus and other device-to-device protocols:

1. Make inter_byte_timeout always be expressed in seconds (allow floating point)
2. Document which systems support `inter_byte_timeout` and its resolution and/or other restrictions

rdpoor:
See also #88 and #373

PR #789, opened 2025-03-13
inter_byte_timeout fix

TheMDev:
Hi, unsure if this will get merged, as it seems the project may no longer be actively maintained. I figured I'd open this to possibly help others experiencing the same issue. This issue seems to date back almost a decade with no fix: https://github.com/pyserial/pyserial/issues/88, https://github.com/pyserial/pyserial/pull/373, https://github.com/pyserial/pyserial/issues/396, https://github.com/pyserial/pyserial/pull/525

The `inter_byte_timeout` argument appears to be correctly set to vtime in the serial class, and the select function does return when the inter byte time exceeds vtime. The `Serial.read()` functions loop only checks timeout, not inter_byte_timeout. This results in the loop continuing until the vmin condition is met, then exit. Simply adding `or self._inter_byte_timeout is not None:` to the break condition resolves this issue.

The `VTIMESerial.read()` function also appears to have an unnecessary loop without an exit condition for any kind of timeout. It was mentioned in https://github.com/pyserial/pyserial/pull/525 that the loop was used to catch errors from the os.read() function. However, this can be accomplished using a raise as it is done in `Serial.read()`, and without the loop, there is no need for an exit condition.

requirements.txt
```pyserial @ git+https://github.com/pyserial/pyserial@0f3e75369707c5246106401f67d3baf361473c8d```


dpgeorge:
Thanks for raising this PR.  This definitely needs to be fixed, and the approach here looks reasonable, although I'd like to test it myself.

TheMDev:
@dpgeorge No worries, I have been running with this patch since I opened the PR and have not had any issues thus far. I was excited to see the project making progress again so I just rebased to fix the merge conflict. I haven't had time to do in depth testing yet on the new release but do not expect there to be any issues. Let me know if you have any questions or concerns, as I would love to get this merged in to help others.

dpgeorge:
@TheMDev can you comment how this PR compares with the approach in #373?

TheMDev:
@dpgeorge It has been a while since I looked at this but, from what I remember the approach in #373 you unnecessarily stay in the loop after the inter byte timeout is exceeded and wait for another character instead of exiting immediately. I believe I noticed this with using long timeouts during testing because it would hang for quite a bit instead of exiting after the inter byte time out was exceeded.

PR #790, opened 2025-03-14
Fix wrong implementation of `in_waiting` in `protocol_socket.py`

### LINKED ISSUE #759: URL Handler Protocol Socket in_waiting is incorrect
The current implementation is shown below from `url_handler.protocol_socket.Serial`. The select call returns whether or not data is ready as shown by the ready loop in the read function and not the number of bytes in waiting. This is particularly a problem when using the `ReaderThread` class as it causes the run loop to "busy" loop and consume excessive amounts of CPU. 

```
@property
def in_waiting(self):
    """Return the number of bytes currently in the input buffer."""
    if not self.is_open:
        raise PortNotOpenError()
    # Poll the socket to see if it is ready for reading.
    # If ready, at least one byte will be to read.
    lr, lw, lx = select.select([self._socket], [], [], 0)
    return len(lr)
```
```

while len(read) < size:
      try:
          ready, _, _ = select.select([self._socket], [], [], timeout.time_left())
```

I have included a possible fix below. I have tested the implementation on Unix (Linux Ubuntu 20.04), but I don't have access to a Windows machine at the moment to verify the windows implementation.

  @property
  def in_waiting(self):
  """Return the number of bytes currently in the input buffer."""
  if not self.is_open:
      raise PortNotOpenError()
      
  if os.name == 'nt':  # Windows
      return len(sock.recv(4096, socket.MSG_PEEK))
      
  else:  # POSIX (Unix-like)
      import fcntl
      import struct
      import termios
      
      buf = fcntl.ioctl(self._socket, termios.FIONREAD, struct.pack("I", 0))
      nbytes = struct.unpack("I", buf)[0]
      return nbytes

yuhaoth:
fix #759 

As the description in #759, the function is wrong. It returns the number of ready sockets. In the context, it will returen 0 or 1. Also the `read()` , the parameter of `recv()` should be available data or left buf size.

### DIFF
--- a/serial/urlhandler/protocol_socket.py
+++ b/serial/urlhandler/protocol_socket.py
@@ -184,7 +184,7 @@ class Serial(SerialBase):
                 # there is nothing to read.
                 if not ready:
                     break   # timeout
-                buf = self._socket.recv(size - len(read))
+                buf = self._socket.recv(min(size - len(read), self.in_waiting))
                 # read should always return some data as select reported it was
                 # ready to read when we get to this point, unless it is EOF
                 if not buf:


projectgus:
@yuhaoth are you still interested in updating this PR?

PR #801, opened 2025-07-08
fix(serial): handle write timeout correctly with overlapped I/O on Wi…

jwzl:
…ndows

- Properly handle `None` for `self._write_timeout` by using `win32.INFINITE`.
- Use `WaitForSingleObject` to wait with timeout before calling `GetOverlappedResult`.
- If timeout occurs (`WAIT_TIMEOUT`), cancel the pending write using `CancelIoEx`.
- Avoid blocking indefinitely on `GetOverlappedResult` by passing `wait=False`.
- Raise `SerialTimeoutException` if the write is incomplete or canceled.

This prevents the serial write from hanging indefinitely when the device becomes unresponsive.

### DIFF
--- a/serial/serialwin32.py
+++ b/serial/serialwin32.py
@@ -318,7 +318,17 @@ class Serial(SerialBase):
 
                 # Wait for the write to complete.
                 #~ win32.WaitForSingleObject(self._overlapped_write.hEvent, win32.INFINITE)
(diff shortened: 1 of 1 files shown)


lbrendel-signum:
@jwzl would a similar but different update need to be done for `read`? I have recently been facing `GetOverlappedResult` blocking on me and have experienced it with read before

PR #803, opened 2025-07-13
Ensure the Windows USB interface number is reset and doesn't persist from previous loop iterations

### LINKED ISSUE #802: iterate_comports on Windows sometimes appends an incorrect interface indicator to the location of USB COM ports
When calling `serial.tools.list_ports.comports()` on Windows, the `ListPortInfo.location` attribute of a USB COM port without multiple interfaces will have the interface number of a previously enumerated port incorrectly applied to it, `':x.0'` for example. This is because the variable `bInterfaceNumber` is set to `None` in the outer _GUID_ loop, and not the inner _device_ loop.

joshstrohminger:
Fixes #802 

`bInterfaceNumber` was only being set to `None` in the outer _GUID_ loop, and not in a deeper loop over each _device_. Because of this, if a USB device didn't have multiple interfaces, but a previously enumerated USB device did have multiple interfaces, the value of `bInterfaceNumber` would still retain the old value and it would be incorrectly applied to the `location` string. The fix simply resets the variable for each COM port.

This is an example of a fictional sequence of devices, showing their locations before and after this fix.

| Order | Hardware ID | Interface | Location Before Fix | Correct | Location After FIx |
| ----- | ----------- | --------- | ------------------- | ------- | ------------------ |
| 1 | USB\VID_1111&PID_1111&MI_00 | 0 | 1-1.1:x.0 | Correct | 1-1.1:x.0 |
| 2 | USB\VID_2222&PID_2222 | None | 1-2.2:x.0 | Wrong | 1-2.2 |
| 3 | USB\VID_3333&PID_3333&MI_01 | 1 | 1-3.3:x.1 | Correct | 1-3.3:x.1 |
| 4 | USB\VID_4444&PID_4444 | None | 1-4.4:x.1 | Wrong | 1-4.4 |
| 5 | USB\VID_5555&PID_5555 | None | 1-5.5:x.1 | wrong | 1-5.5 |

### DIFF
--- a/serial/tools/list_ports_windows.py
+++ b/serial/tools/list_ports_windows.py
@@ -259,7 +259,6 @@ def iterate_comports():
 
     # repeat for all possible GUIDs
     for index in range(len(GUIDs)):
-        bInterfaceNumber = None
         g_hdi = SetupDiGetClassDevs(
             ctypes.byref(GUIDs[index]),
             None,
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thank you for the fix

PR #810, opened 2025-10-30
force kill serial port

marcusnk237:
The user have the possibilty during the Serial Port object creation, to force kill any previous instance of the targetted Serial Port. This option is useful for microcontrollers applications involving Serial Port communications.

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -190,13 +190,14 @@ class SerialBase(io.RawIOBase):
                  dsrdtr=False,
                  inter_byte_timeout=None,
                  exclusive=None,
+                 forced = False,
                  **kwargs):
         """\
         Initialize comm port object. If a "port" is given, then the port will be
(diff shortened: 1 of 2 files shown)


marcusnk237:
The user have the possibilty during the Serial Port object creation, to force kill any previous instance of the targetted Serial Port. This option is useful for microcontrollers applications involving Serial Port communications.

projectgus (commented):
This is an impressive hack, I would never have thought to walk all live objects in the interpreter and look for existing Serial instances!

That said, I don't think this is an approach that we want to try and maintain in pyserial itself into the future - it's too much of a hack, sorry.

The good news is that I don't think this has to be inside pyserial to work? It looks like anyone could add a routine in their app that loops through all the live objects and calls `.close()` on the port (and/or calls the win32 APIs to forcible give up the handle).

PR #812, opened 2025-10-31
Fix Read the Docs builds

kurtmckee:

* Add a `.readthedocs.yaml` config file. This file is now required by Read the Docs.
* Add a `documentation/requirements.txt` file. This allows the docs to be built locally.
* Configure Sphinx to use the Read the Docs theme. This ensures that the local and Read the Docs builds match.

When this merges, [Read the Docs builds should start passing](https://app.readthedocs.org/projects/pyserial/builds/). I'll monitor to confirm that this is the case, though.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,15 @@
+version: 2
+
+build:
+  os: 'ubuntu-24.04'
+  tools:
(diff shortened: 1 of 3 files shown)


zsquareplusc:
cool, thank you

PR #811, opened 2025-10-30
Migrate to a `pyproject.toml` file for package metadata

kurtmckee:
Hello @zsquareplusc! This PR introduces the following changes:

* Migrate package metadata to a standard `pyproject.toml` (PEP 621)
* Drop support for EOL Python versions (Python 3.9 and lower)
* Remove references to `setup.py` throughout the docs

The project can now be built using the standard `build` package:

```
python -m pip install build
python -m build
```

These changes resolve the following warnings
that were thrown when building the project:

```
SetuptoolsDeprecationWarning: License classifiers are deprecated.
!!

********************************************************************************
Please consider removing the following classifiers in favor of a SPDX license expression:

License :: OSI Approved :: BSD License

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
********************************************************************************

!!
```

```
SetuptoolsDeprecationWarning: bdist_wheel.universal is deprecated
!!

********************************************************************************
With Python 2.7 end-of-life, support for building universal wheels
(i.e., wheels that support both Python 2 and Python 3)
is being obviated.
Please discontinue using this option, or if you still need it,
file an issue with pypa/setuptools describing your use case.

This deprecation is overdue, please update your project and remove deprecated
calls to avoid build errors in the future.
********************************************************************************

!!
```

Followup work is required to remove references to old Python versions.

PR #813, opened 2025-10-31
Add a tox config, and run test suite in CI

kurtmckee:

* Locally test the project on all supported Python versions using tox.
* Run the test suite on Linux and Windows in CI.
  _macOS testing freezes midway through testing. This can be investigated in subsequent work._
* Generate an HTML coverage report to show what code is tested.
* Escalate all test suite warnings to errors.
* Remove references to Travis CI.

[Example CI run](https://github.com/kurtmckee/pr-pyserial/actions/runs/18959268347)

The test suite can be run locally using tox (the tox-uv plugin is suggested for speed):

```
python -m pip install tox tox-uv
tox
```

The tests can be run in parallel, too, which is significantly faster:

```
tox p
```

Once merged, GitHub will automatically run CI against new and updated pull requests, which can help build confidence in the quality of PRs.


kurtmckee:
@projectgus Great questions.

> wondering if we could get a similar result with `uv run`?

It's impossible to use `uv run` to accomplish what this PR is introducing. `uv run` is able to install dependencies in a virtual environment, same as tox, but at the end of the day it simply runs a script. tox creates virtual environments, too, but can coordinate:

* testing against multiple version of Python at the same time _(in this case, Python 3.10+)_
* ordering the dependent execution of tox environments _(like first collecting coverage reports from all Python versions _and only after that's done_ generating a combined coverage report)_
* running necessary setup commands _(like `coverage combine`)_

> I'm not that familiar with `tox` ... Are there particular benefits to using a stack with `tox` these days

Yes.

> I notice also pulls in `poetry`

That's not strictly necessary, and I can remove that if desired. However, I strongly recommend against it.

poetry is introduced only for the test suite dependencies. It's the only tool I've found so far that can export a cross-platform `requirements.txt` file _that contains markers for all platforms_, which ensures that the test suites are running with reproducible, consistent dependencies across all platforms.

The boilerplate you're seeing has these technical advantages:

* Reproducible, locked test suite dependencies across all platforms
* A trivial method to update those dependencies (`tox -m update`)
* What you run on your machine matches what I run on my machine, and we both match CI

The risk of removing those dependencies is serious, particularly for a project like pyserial that I'm trying to get momentum behind. In particular, non-locked test suite dependencies means that PRs from future contributors _might fail simply because pytest released a new minor version_.

> Popping a comment here to follow this up at some point

No need. :grin: I've got that -- and more changes -- in the hopper!

However, tox, CI, and the package metadata builds have to merge first to ensure that I have the beginnings of a baseline to work against.

zsquareplusc:
the general direction looks fine

i've used pytest in other projects as it also integrates in e.g. VSCode test plugin. at least for running the tests on the current configuration or venv. i might have added a `pytest.ini`.

tox seems to be useful to get the test matrix over OS and py versions, especially in CI.

the requirements dir and poetry look complex :) but OK could start with this and fine tune later.

the tests can also be run against real hardware by passing in the port as arg on command line if test test* scripts are run directly or by patching PORT global var (maybe there would also be a better way to pass in this as a test parameter). i would say it's normally not needed to run this across multiple python versions but multiple OS are useful. but we currently do not have CI servers with real hardware.

on linux some of the tests might run with pty's or channels created with `socat`. on windows com0com driver or similar projects also support virtual serial ports connected to each other. but probably not possible to install drivers within CI. those solutions create two ends so the tests would need a second process echoing the data on the second port or tests specifically that open two ports.

testing with the virtual serial ports would have the advantage to use the "real" `Serial` class for each platform and not the loop:// implementation

projectgus:
@kurtmckee Thanks for detailed explanation, that makes sense. The tight integration with a coverage tool is particularly neat!

kurtmckee:
_Rebased on `master`, which resolved the failing `build` test._

@zsquareplusc 

> i might have added a `pytest.ini`

Now that #811 has merged and I don't have to worry about merge conflicts, I've moved the coverage config to `pyproject.toml` and have added a pytest config there as well. pytest is ratcheted down to escalate all warnings to errors.

> it's normally not needed to run this across multiple python versions

I strongly disagree in the context of evaluating others' PRs. It is not uncommon for Python to introduce aliases in newer Python versions, or to deprecate things that will eventually be removed in a coming Python version. Testing across Python versions helps ensure that a PR doesn't introduce a new reference that doesn't exist in older Python versions.

I have receipts for that claim: Python deprecated calls to `datetime.datetime.utcnow()` starting in Python 3.12. It has become common to see PRs from people who are wanting to fix `DeprecationWarning` in their test suites...but their PRs convert such calls to `datetime.datetime.now(tz=datetime.UTC)`.

However, `datetime.UTC` is an alias for `datetime.timezone.utc` that only exists in Python 3.11 and higher. Python 3.10 will throw an `AttributeError`. You wouldn't see that failure unless you were testing all supported Python versions.

(And to your point regarding having a pytest config, you wouldn't see the `DeprecationWarning` unless you were testing Python 3.12 and higher, _and_ with pytest configured to escalate warnings to errors...and that's how it's configured.)

Thank you all for your reviews of this PR! If you have specific requests for changes, please let me know _(edit: please phrase it as "please change xyz" so that it's very clear what's being requested)_. Otherwise, it sounds like this is regarded as a helpful change, and since my further work on the project depends on having strong testing I'm hopeful that this can be merged soon.

zsquareplusc:
regarding testing across multiple Python versions, i specifically meant when running the tests against real hardware.
of course running the test suite matrix fully can be done in CI (and locally). but it is different when you run *additional* tests against hardware.

ideally pure software tests would detect any incompatibilities with python versions or OS. so running additional tests with hardware would be more about the hardware and driver mysteries and then running all py combinations is not a bad thing, but it is high test effort versus a small chance to find a new bug regarding python versions.

but as mentioned earlier, the current tests are run with the loop:// "port", which is a pure python implementation and therefore does not cover much of the code of real the `Serial` class of each platform. it's imported but not used. if we want to improve the test coverage of CI tests, it would be good if the real `Serial` class could be used to open/read/write.

i've tried running tests with socat using this helper https://github.com/pyserial/pyserial/blob/feat-loop-test/test/loopback.py some fraction of the tests fail because a /dev/pts is not as capable as a real /dev/tty but many tests also work

kurtmckee:
> i specifically meant when running the tests against real hardware

Excellent distinction, and I agree.

I apologize if I'm being dense, but I'm not clear if you're suggesting the scope of this PR should increase to do an overhaul of the existing testing. Is this a discussion about a blocking issue for merging this PR?

I'd prefer that this PR not expand in scope. However, there's plenty of testing work that can be done after this PR merges.

zsquareplusc:
OK merged, though i'm not yet very familiar with some of the tools used and it adds numerous dependencies (which to some extent is of course OK for testing, i'd like to keep the dependencies for pySerial itself for the user of the library minimal)

PR #814, opened 2025-11-05
Remove support for Jython and IronPython

kurtmckee:

* Remove support for Jython.
  Jython only supports Python 2.x syntax.

* Remove support for IronPython.
  IronPython only supports Python 2.7 and 3.4 syntax.


zsquareplusc:
Agreed these platforms are probably not in wide use and the restriction on syntax may hold us back.
Users can still install old releases up to 3.5.

PR #815, opened 2025-11-07
Test and update `serial/tools/hexlify_codec.py`

kurtmckee:
> [!NOTE]
>
> Two commits are introduced here, and it's recommend to review them in order.
>
> * 50c42ad5e8a2d31bead353c358fb54944b7b72d0 adds tests that execute every line in `hexlify_codec.py`. Doing so revealed a critical failure in the code that required changes to the `Codec.encode()` and `Codec.decode()` return values. The fixes are included in this commit.
> * 691f33cc94268ee68907ab8ec606e6d5051ecd8f updates `hexlify_codec.py` to use Python 3 syntax and features.

----

This PR introduces the following changes:

*   Test `serial/tools/hexlify_codec.py` to 100%
    
    Testing revealed a critical failure in the return types
    for the `Codec.encode()` and `Codec.decode()` methods.
    This is fixed by also returning the length of `data`.
    
    In addition, the state is retrieved and stored
    using the `.getstate()` and `.setstate()` methods,
    which allowed testing to reach 100%.

* Remove Python 2 compatibility code.
* Replace custom code with `bytes.fromhex()` and `.hex()` calls.
* Add type annotations.
  (These have been cursorily checked by mypy
  but are not yet guaranteed to be correct.)
* Run the file through pyupgrade/isort/black/ruff.
    
The test suite continued to pass with no other changes,
and running miniterm with the RX encoding set to "hexlify"
printed pairs of uppercase, space-separated hex digit pairs
as expected.

### DIFF
--- a/serial/tools/hexlify_codec.py
+++ b/serial/tools/hexlify_codec.py
@@ -1,77 +1,68 @@
-#! python
-#
-# This is a codec to create and decode hexdumps with spaces between characters. used by miniterm.
+# This is a codec to create and decode hexdumps with spaces between characters.
 #
 # This file is part of pySerial. https://github.com/pyserial/pyserial
 # (C) 2015-2016 Chris Liechti <cliechti@gmx.net>
 #
 # SPDX-License-Identifier:    BSD-3-Clause
 """\
-Python 'hex' Codec - 2-digit hex with spaces content transfer encoding.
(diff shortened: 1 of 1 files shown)


zsquareplusc:
Thanks. I see you have different preference for quotes, " vs '

PR #818, opened 2025-11-10
SerialBase: set logger = None in base class constructor and nowhere else

cdleonard:
Many child class implementations individually set logger=None inside their constructor or open() methods. Having this done only once in the base class constructor is simpler and also useful:

You can now always set the logger between `__init__` and `open` and get useful logging output from initialization code.

This is a more general version of
commit b63d94fd4e85 ("rfc2217: do not set logger = None inside open()") in #764

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -200,6 +200,10 @@ class SerialBase(io.RawIOBase):
         self.is_open = False
         self.portstr = None
         self.name = None
+
+        # logging disabled by default but can be set to a Logger instance after __init__
+        self.logger = None
+
         # correct values are assigned below through properties
         self._port = None
         self._baudrate = None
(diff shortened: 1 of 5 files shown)


serial/serialutil.py
```diff
@@ -217,6 +217,7 @@ def __init__(self,
         self._dtr_state = True
         self._break_state = False
         self._exclusive = None
+        self.logger = None
```
kurtmckee: This line appears in a list of attributes that are assigned via properties. It should move up above the comment `correct values are assigned below through properties`.

kurtmckee:
@zsquareplusc @andrewleech Please merge this PR. It standardizes the logger initialization across the `SerialBase` classes.

PR #819, opened 2025-11-11
Update URLs to match current redirects

kurtmckee:

* URLs that redirect have been updated.
* Some unadorned URLs in the documentation are updated to use RST embedded link syntax.
* A redirect that occurred when building the documentation has been addressed in `documentation/conf.py`.

The following message was displayed when building the docs, and this is fixed as a part of this change:

```
loading intersphinx inventory 'py' from https://docs.python.org/objects.inv ...
intersphinx inventory has moved: https://docs.python.org/objects.inv -> https://docs.python.org/3/objects.inv
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -945,7 +945,7 @@ v1.0 - 2002-02-13
 =================
 
 - First public release.
-- Split from the pybsl application (see http://mspgcc.sourceforge.net)
+- Split from the pybsl application (see https://sourceforge.net/projects/mspgcc/)
 
 **New Features**
 
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -36,4 +36,4 @@ Individual files contain the following tag instead of the full license text.
     SPDX-License-Identifier:    BSD-3-Clause
 
 This enables machine processing of license information based on the SPDX
-License Identifiers that are here available: http://spdx.org/licenses/
+License Identifiers that are here available: https://spdx.org/licenses/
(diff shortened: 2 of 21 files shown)


kurtmckee:
_Rebased on `master` and fixed merge conflict._

kurtmckee:
@dpgeorge @projectgus Friendly ping to review and merge this. I'm concerned about merge conflicts for additional work that needs to happen in the README and documentation.

kurtmckee:
@zsquareplusc @andrewleech Ping! I'd like to do additional work on the README and other documentation to prep for the v3.6 release, but I'm blocked by concerns of merge conflicts with this existing PR.

projectgus (approved):

projectgus:
Thanks @kurtmckee. Out of interest, when you ran this update (that I assume used a script?) you didn't run into any URIs that no longer exist at all?

kurtmckee:
@projectgus Unfortunately it was done by hand, and I _did_ encounter several URLs that no longer resolve. I didn't touch those, as some appeared to be attribution references, but at the time I was working on this archive.org was simultaneously down so I wasn't able to look for a secondary URL that would resolve.

(I didn't use a script in part because I'm only familiar with Sphinx extensions and tools that consume HTML, and there are a number of URLs in code comments.)

andrewleech:
Thanks for this cleanup, I did try to automatically check the urls's you've changed and they all look good.

Even if all the links can't be perfectly fixed, this is a great improvement!

PR #820, opened 2025-11-11
Fix `UnboundLocalError`

### LINKED ISSUE #617: UnboundLocalError when using PosixPollSerial with timeout

kurtmckee:
This change supersedes #618 by merging `master` into the branch and incorporating feedback from #618.

It differs slightly from the feedback, however: the `break` at line 834 is preserved so that the `for` loop exits immediately if `fd == self.pipe_abort_read_r`.

@dpgeorge If you intentionally left the `break` out of your suggested change, I can remove it here. Note that the test suite currently does not touch this code.

Fixes #617

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -825,21 +825,19 @@ class PosixPollSerial(Serial):
         poll.register(self.pipe_abort_read_r, select.POLLIN | select.POLLERR | select.POLLHUP | select.POLLNVAL)
         if size > 0:
             while len(read) < size:
-                fd = None #In case poll.poll() returns nothing
-                # print "\tread(): size",size, "have", len(read)    #debug
                 # wait until device becomes ready to read (or something fails)
+                abort = False
                 for fd, event in poll.poll(None if timeout.is_infinite else (timeout.time_left() * 1000)):
                     if fd == self.pipe_abort_read_r:
+                        os.read(fd, 1000)
+                        abort = True
                         break
                     if event & (select.POLLERR | select.POLLHUP | select.POLLNVAL):
                         raise SerialException('device reports error (poll)')
-                    #  we don't care if it is select.POLLIN or timeout, that's
-                    #  handled below
-                if fd == self.pipe_abort_read_r:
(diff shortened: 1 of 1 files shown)


dpgeorge:
> It differs slightly from the feedback, however: the `break` at line 834 is preserved so that the `for` loop exits immediately if `fd == self.pipe_abort_read_r`.

What you have here is correct.  My suggestion should have kept that `break`.

BTW, I think it's OK to rebase the commit from #618, instead of adding a merge commit.

kurtmckee:
If it were my own work I would typically rebase. However, rebasing strips the original signatures, so I don't do that to others' work. For example, all of my commits to this repository are attributed to me but unverified because of how they were committed.

> image

I prefer merge commits so that signatures are preserved, but squash-merges will still be signed by GitHub so that they appear to be verified in GitHub's web interface.

kurtmckee:
@dpgeorge I've eliminated the merge commit. When CI checks out a PR branch for testing, `actions/checkout` auto-merges with `master` by default. Rebasing and merging is unnecessary because CI handles it for us! :partying_face: 

For example, the CI run that just completed merged commit `fc9d282` with `master` (`a829792`) [[reference](https://github.com/pyserial/pyserial/actions/runs/19253915911/job/55044458181?pr=820#step:2:82)]:

```
HEAD is now at 651ad7d Merge fc9d282c2de7da238d789d587993e5c410392bf0 into a8297923588203fd0de5545a333320f710ff6361
```

dpgeorge (approved):

PR #821, opened 2025-11-11
Remove all deprecated interfaces

kurtmckee:

* Remove `.nonblocking()` (POSIX only).
* Remove `.portstr` (use `.name`).
* Remove `.inWaiting()` (use `.in_waiting`).
* Remove `isOpen()` (use `.is_open`).
* Remove `.writeTimeout` (use `.write_timeout`).
* Remove `.interCharTimeout` (use `.inter_byte_timeout`).
* Remove `.sendBreak` (use `.send_break`).
* Remove `.flushInput()` (use `.reset_input_buffer()`).
* Remove `.flushOutput()` (use `.reset_output_buffer()`).
* Remove `.setBreak()` (use `.break_condition()`).
* Remove `.setRTS()` (use `.rts`).
* Remove `.setDTR()` (use `.dtr`).
* Remove `.getCTS()` (use `.cts`).
* Remove `.getDSR()` (use `.dsr`).
* Remove `.getRI()` (use `.ri`).
* Remove `.getCD()` (use `.cd`).
* Remove `.getSettingsDict()` (use `.get_settings()`).
* Remove `.applySettingsDict()` (use `.apply_settings()`).
* Remove `.outWaiting()` (use `.out_waiting`).
* Remove `.setXON()` (use `.set_output_flow_control()`).
* Remove `.flowControlOut()` (use `.set_input_flow_control()`).
* Remove `.rtsToggle` (Windows only; use `.rs485_mode`).

In addition:

* Remove the `writeTimeout` initialization parameter. Use `write_timeout` instead.
* Remove the `interCharTimeout` initialization parameter. Use `inter_byte_timeout` instead.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -837,3 +837,32 @@ Removed:
 
 - Remove support for Jython.
 - Remove support for IronPython.
+- Remove all deprecated methods and attributes.
+
+  - Remove ``.nonblocking()`` (POSIX only).
+  - Remove ``.portstr`` (use ``.name``).
+  - Remove ``.inWaiting()`` (use ``.in_waiting``).
+  - Remove ``isOpen()`` (use ``.is_open``).
+  - Remove ``.writeTimeout`` (use ``.write_timeout``).
+  - Remove ``.interCharTimeout`` (use ``.inter_byte_timeout``).
+  - Remove ``.sendBreak`` (use ``.send_break``).
+  - Remove ``.flushInput()`` (use ``.reset_input_buffer()``).
+  - Remove ``.flushOutput()`` (use ``.reset_output_buffer()``).
+  - Remove ``.setBreak()`` (use ``.break_condition()``).
+  - Remove ``.setRTS()`` (use ``.rts``).
+  - Remove ``.setDTR()`` (use ``.dtr``).
+  - Remove ``.getCTS()`` (use ``.cts``).
+  - Remove ``.getDSR()`` (use ``.dsr``).
+  - Remove ``.getRI()`` (use ``.ri``).
+  - Remove ``.getCD()`` (use ``.cd``).
+  - Remove ``.getSettingsDict()`` (use ``.get_settings()``).
+  - Remove ``.applySettingsDict()`` (use ``.apply_settings()``).
+  - Remove ``.outWaiting()`` (use ``.out_waiting``).
+  - Remove ``.setXON()`` (use ``.set_output_flow_control()``).
+  - Remove ``.flowControlOut()`` (use ``.set_input_flow_control()``).
+  - Remove ``.rtsToggle`` (Windows only; use ``.rs485_mode``).
+
+- Remove all deprecated initialization parameters.
+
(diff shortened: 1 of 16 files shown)


dpgeorge:
I'm generally in favour of this, but I suggest a wider discussion with @zsquareplusc, as well as @andrewleech and @projectgus .  In particular using semver and probably the next release would be 4.0.0.

kurtmckee:
Agreed. `git blame` showed that some of the deprecations were introduced 9 years ago (back in 2016), so hopefully this is noncontroversial.

projectgus:
Suggest we proceed as follows:

* Release a v3.6 with the changes merged in the five years since v3.5, plus (ideally) the glibc fix in #808 
* Merge this PR
* Make the next release v4.0

If this sounds amenable then I can set up a couple of Milestones in the project to make it easier to track.

dpgeorge:
> Suggest we proceed as follows:
> ...

I agree with your proposal (except the v4 release would be v4.0.0 for semver).

The only things since v3.5 that would warrant a major version bump are the removal of Python 3.9 and lower, and Jython and IronPython support.  But I think it's fine to just drop those platforms, their support will just end at v3.5 and other platforms continue the v3 series with v3.6 and are not impacted.

kurtmckee:
@projectgus I recommend to skip milestones and to start cutting releases. Milestones increase the likelihood that no releases actually get published, and Chris is the only person who can publish to PyPI, regardless of what administrative work is invested in drawing up milestones.

@zsquareplusc Would you be willing to publish v3.6 to PyPI with the code that's merged to `main`? I would also like to request that you provide PyPI and Read the Docs permissions to the others to reduce stagnation risk in the project.

andrewleech:
Removing these should help clarify the code and make it easier to maintain in future, so is a good move. It _will_ certainly break a lot of third party code so I definitely agree it needs to be a major release. 

I think there should be an upgrade guide page that lists all these and the replacement api's for them to serve as a single source to assist users in their migrations.

kurtmckee:
@andrewleech That's a good suggestion! I can write that up and update this PR once the code in `master` has been published in a new release.

@zsquareplusc Would you be willing to publish v3.6? I can update the CHANGELOG and bump the project metadata if you're willing to publish a new release after it merges, and provide start-to-finish commands and instructions if it's helpful. :+1: 

Let me know either way! :grin:

zsquareplusc:
I would not remove those yet. pySerial will be in use industrial settings where 10 and more years old code is still running and removing those will break user code. The old method names on the Serial object are trivial redirects to the new names it's no effort to "maintain" these a bit longer. Maybe in a next step, the `warning` module could be used to make the deprecation visible at runtime, until now it was only docs.

i see a lot of other good updates in this PR, like updating all the internal usage and tests that those only use the new names. i'd just keep the redirects for a little while longer.

kurtmckee:
@zsquareplusc Super, I'll close this in favor of simply updating the internal usage to avoid the deprecated names.

Issue #625, opened 2021-11-18
Calling .flush() hangs forever - when to use .flush()?


benthorner:
Code to reproduce the problem:

```
import pty, os, time
master, slave = pty.openpty()

def loop():
    while True:
        data = os.read(master, 1)
        print(data)  # <--- prints nothing
        time.sleep(.1)

from threading import Thread
thread = Thread(target=loop, daemon=True)
thread.start()

from serial import Serial
serial = Serial(os.ttyname(slave))

serial.write(b'123')
serial.flush()  # <--- gets stuck here

time.sleep(3)  # allow thread to read all data
```

If you comment out `serial.flush()`, the program runs to completion and all 3 bytes are printed out. When I manually quit the program with a Ctrl-C, the stacktrace indicates it's hanging on [`termios.tcdrain(self.fd)`](https://github.com/pyserial/pyserial/blob/5ce1773fdce16a0b184363695a7d7d9483c4da0c/serial/serialposix.py#L680) in `.flush()`.

I tried looking up `tcdrain` but can't find much information about it. Nothing I've read explains why the call to `.flush()` would hang forever, or why it would interfere with the other thread reading the data that's already been written.

The only thing that looked relevant is [the GIL](https://wiki.python.org/moin/GlobalInterpreterLock), but even that seems far fetched.

Does anyone know why this is happening? 

More generally, is there some guidance about when `.flush()` is necessary?

benthorner:
@OscarL yes, I forgot to mention I'm running this on a Mac.

That's useful to know someone else is experiencing the same issue, albeit in a different context. I also see what you mean about the Python vs Termios implementations of `tcdrain`:

- [Python docs say it will wait](https://docs.python.org/3/library/termios.html#termios.tcdrain)
- [Termios docs says it will discard](https://man7.org/linux/man-pages/man3/termios.3.html)

I notice [the Python implementation just calls the native system function](https://github.com/python/cpython/blob/main/Modules/termios.c#L259), which has the same "wait" description of its behaviour in [the Mac docs](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/tcdrain.3.html), although I'm not sure if it's relevant.

It seems like this is a Mac-specific issue at the C language level and needs raising with Apple, although I'm not sure as it's so low-level. I could try and post about it in the WDC forum. What do you think?

benthorner:
[...]

PR #817, opened 2025-11-10
Thoroughly test `serial/tools/list_ports.py` and fix a crash bug

kurtmckee:
> [!NOTE]
>
> This PR is broken into two logical changes with a single commit each. It is recommended to review the changes one commit at a time.
>
> 1. 4ac0034e10471e61bb51d1c79097f01dd85328bd tests `serial/tools/list_ports.py` to 100% coverage.
>
>    This revealed a critical bug in the `-n` option's logic, which would result in a `NameError` in most cases. The bug is fixed in this commit.
>
> 2. a1167b2c0c96149bc5656ef27019b053a90d416f updates the Python syntax and adds type annotations. Issues reported by mypy were resolved.

This PR introduces the following changes:

*   Test `serial/tools/list_ports.py` to 100%
    
    Testing revealed that using the `-n` option
    would typically result in a `NameError` due to a typo.
    This is fixed by referencing `args.n`.
    
    In addition, unreachable code is now comment-ignored
    for coverage purposes.

* Add type annotations.
    
  This has been run cursorily through mypy,
  and reported issues have been resolved.
  However, type annotations are not guaranteed to be correct yet.
    
* Run the code through pyupgrade/isort.
    
  Based on past feedback, the code is not run through black/ruff
  because these will convert quotation marks.

With these changes, code coverage increases from ~25% to ~29%.

PR #822, opened 2025-11-11
Re-enable macOS CI to demonstrate PTY test hang

andrewleech:
## Re-enable macOS CI to demonstrate PTY test hang

The macOS runner was commented out because tests were freezing. This PR re-enables it to demonstrate the issue and start discussion on the best fix.

### What I Found

The hang happens in `test_pty_serial_read` when it calls `slave.flush()`. This triggers `tcdrain()` which has different behavior on macOS vs Linux when dealing with PTY devices:

- **Linux**: `tcdrain()` waits until data leaves the output buffer
- **macOS**: `tcdrain()` waits until someone *reads* the data from the other end

Since the test writes from the slave side before setting up a reader on the master, macOS blocks forever. This is a documented macOS kernel quirk (see pyserial/pyserial#625, python/cpython#97001).

### Potential Fixes

Two approaches to discuss:

#### Option 1: Fix the Test
Remove `slave.flush()` in `test_pty_serial_read` - it's not actually needed to verify data transmission.

**Pros:**
- Minimal change
- Test still validates PTY communication
- Zero risk to library behavior

**Cons:**
- Users calling `flush()` on macOS PTYs will still hit this
- Just papers over the library issue

#### Option 2: Fix the Library  
Make `flush()` detect PTY devices on macOS and skip the blocking `tcdrain()` call.

**Pros:**
- Fixes it for everyone using PTYs on macOS
- Matches Linux behavior
- Helps real use cases (terminal emulators, test frameworks, etc.)

**Cons:**
- More invasive change
- Need careful PTY detection logic
- Could affect semantics for edge cases

### This PR

This just re-enables the macOS runner so we can see the hang in CI and discuss which approach makes sense. The tests *will* fail/timeout - that's intentional to demonstrate the issue.

@zsquareplusc thoughts on which direction to take? I've got test branches for both approaches which fixes the problem but I'm not sure if there are "legitimate" usages of pty with the library outside of the unit test suite here that warrants fixing it in the library itself?

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -678,9 +678,42 @@ class Serial(SerialBase, PlatformSpecific):
         """\
         Flush of file like objects. In this case, wait until all data
         is written.
+
+        On macOS where the serial device corresponds to a PTY, the underlying
+        call to termios.tcdrain() can block indefinitely because it waits for
+        data to be read from the other end, which may never happen. So, for
+        PTYs, we skip tcdrain() since data is already in the kernel buffer
+        after write().
         """
         if not self.is_open:
             raise PortNotOpenError()
+
+        # On macOS, tcdrain() can block indefinitely on PTY devices.
+        # See: https://github.com/pyserial/pyserial/issues/625
+        #      https://github.com/python/cpython/issues/97001
+        if sys.platform.startswith('darwin'):
+            try:
+                dev_name = os.ttyname(self.fd)
+                # macOS PTY slave devices are /dev/ttys*
+                # BSD-style PTY devices are /dev/pty[pqrs]*
+                pty_device_names = (
+                    '/dev/ttys',
+                    '/dev/ptyp',
+                    '/dev/ptyq',
+                    '/dev/ptyr',
+                    '/dev/ptys',
+                )
+                if dev_name.startswith(pty_device_names):
+                    # For PTY devices on macOS, tcdrain() blocks indefinitely
+                    # because it waits for data to be read from the other end,
+                    # which may never happen. Since the data is already in the
+                    # kernel buffer after write(), we consider flush() complete.
+                    return
+            except OSError:
+                # If we can't determine the device name, fall back to tcdrain.
+                # This is safe because real serial ports won't fail ttyname().
+                pass
+
         while True:
             try:
                 termios.tcdrain(self.fd)


projectgus:
FWIW Option 2 sounds reasonable to me. Thanks for digging into this, @andrewleech!

kurtmckee:
@andrewleech I've implemented an "option 3" for expediency in #827. It allows the macOS test suite to pass by putting the call to `.read()` in a thread so that the call to `.tcdrain()` can succeed.

It may be ideal to pursue option #2 later macOS needs to be enabled in CI and #827 makes that happen without removing the call to `.flush()`.

Thanks for hunting down the root cause of this issue! :heart: :rocket:

andrewleech:
I realized fixing the library is important, even just for the use case of pyserial based tools communicating with qemu which uses pty's.

I've got a working fix for this on my Mac, I just don't use it very often and it's got a flat battery so I haven't got back to it to push it to a PR!

kurtmckee:
Excellent! I don't have a Mac so I need macOS running in CI to continue hammering on the test suite. Let's tackle this from both sides!

I'd like to see #827 and #817 get merged so I can continue the work of expanding the test suite to cover more code. :pray: :chart_with_upwards_trend:

andrewleech:
Finally got my Mac charged up! Just pushed the library fix.

The approach: detect PTY devices on macOS (by checking if the path starts with `/dev/ttys*` or `/dev/pty[pqrs]*`) and skip the `tcdrain()` call entirely. Once `write()` completes, the data is in the kernel buffer, which is good enough for PTY semantics. Real serial hardware still gets the normal `tcdrain()` behavior.

This should fix the issue for anyone using pyserial with PTYs on macOS - not just our tests, but also folks communicating with QEMU or other emulators.

@kurtmckee your threading fix in #827 is a reasonable workaround for the test specifically, but this library fix means we don't need to restructure the test at all - the original test code with `flush()` now just works. Happy to discuss which approach (or both?) makes sense to merge.

serial/serialposix.py
```diff
@@ -674,9 +674,38 @@ def flush(self):
...
+                          dev_name.startswith('/dev/ptyq') or
+                          dev_name.startswith('/dev/ptyr') or
+                          dev_name.startswith('/dev/ptys'))
+                if is_pty:
+                    # For PTY devices on macOS, tcdrain() blocks indefinitely
+                    # because it waits for data to be read from the other end,
+                    # which may never happen. Since the data is already in the
+                    # kernel buffer after write(), we consider flush() complete.
+                    return
+            except (OSError, IOError):
```
kurtmckee: `IOError` is an alias for `OSError`.

```suggestion
            except OSError:
```

kurtmckee:
pyserial has code that checks for device symlinks when listing available ports. Is there ever a chance that sniffing for the device name could return a symlink instead of the underlying port?

Also, this currently has merge conflicts so CI isn't running against this branch yet.

andrewleech:
> pyserial has code that checks for device symlinks when listing available ports. Is there ever a chance that sniffing for the device name could return a symlink instead of the underlying port?

I'm pretty sure `dev_name = os.ttyname(self.fd)` should be reliable here, it should return the real device path associated with the file descriptor, not the path that was used to open it.

I've just rebased, hopefully it works properly now!

dpgeorge:
Reading through the linked issues for macOS and CPython, it seems that the root cause is a bug in CPython, where it does not release the GIL when doing a `tcdrain()`.  And it just so happens that Linux is OK with this but macOS is not.

It seems that CPython fixed this bug in https://github.com/python/cpython/pull/99503, which is back ported as far as CPython 3.10.

So... the work around added in this PR should not be needed for CPython 3.10 and above.  And the CI here uses CPython 3.10.11 and above.

Therefore, I'm confused what the real issue is.  Is there really still a bug in macOS kernel with `tcdrain()`?

kurtmckee:
@dpgeorge The bug you linked to fixed a full-on interpreter lock caused by the failure to release the GIL. It didn't resolve the macOS kernel requirement that something read from the PTY before `tcdrain()` returns.

As demonstrated in #827, it's possible in CPython 3.10+ to have a secondary thread read from the PTY while the main thread calls `tcdrain()`, since 3.10+ now releases the GIL so that the reader thread is able to run at all.

The test suite fix in #827 would apparently have still hung on CPython 3.9, which I guess still locks the GIL and blocks all other threads.

andrewleech:
> Therefore, I'm confused what the real issue is. Is there really still a bug in macOS kernel with `tcdrain()`?

There's definitely an issue still yes; the initial actions run for this PR failed in the same way I saw locally, with the test suite locking up during the test that used pty tcdrain.
https://github.com/pyserial/pyserial/actions/runs/19255227996/job/55048419958

Screenshot_20251205-072608

This first test run was with python 3.14.

@kurtmckee that sounds like a reasonable suggestion to the cause.  I've tried to use some AI research to deep dive this; came up with: https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/bsd/kern/tty.c#L1657

  Confirmed: macOS PTY tcdrain() Behavior
  The XNU kernel's ttywait() function (called by TIOCDRAIN ioctl) waits until:
  1. tp->t_outq.c_cc (output queue character count) is zero, AND
  2. TS_BUSY flag is cleared

  For PTY devices specifically, the output queue doesn't drain until the master side reads the data. This differs from real serial hardware where data leaves the queue when transmitted to the wire.

  Sources Confirming This:
  1. https://github.com/pyserial/pyserial/issues/625 - Directly documents:
  "macOS interprets tcdrain() differently than Linux when dealing with PTY devices...macOS: tcdrain() waits until data is actually read from the other PTY endpoint"
  2. https://github.com/python/cpython/issues/97001 - The original CPython issue confirms the hang occurs specifically on macOS with PTYs, working fine on Linux.
  3. https://developer.apple.com/forums/thread/715119 - Documents the macOS-specific behavior (though the thread content wasn't fully extractable).
  4. https://github.com/apple/darwin-xnu/blob/main/bsd/kern/tty.c - Shows ttywait() loops while tp->t_outq.c_cc has data. For PTYs, this queue empties only when the master reads.

dpgeorge:
OK, thanks for the extra info!

PR #823, opened 2025-11-11
Always store `OSError.errno` if available

### LINKED ISSUE #215: errno not consistently set when raising SerialException
Occasionally, when a `SerialException` is raised it does not contain an `errno`, instead it is stringified and passed as the `message` of the exception.

This makes the calling code less elegant because it has to try to parse the text of the error in order to figure out what happened.

Here is an example: https://github.com/pyserial/pyserial/blob/master/serial/serialposix.py#L508

And another one, where an `errno` is passed: https://github.com/pyserial/pyserial/blob/master/serial/serialposix.py#L268

kurtmckee:
Exactly what it says on the tin.

Closes #215


dpgeorge (approved):
Looks good, thanks!

(Note to other reviewers: the form `SerialException(errno, msg)` is already in use elsewhere in pyserial, so this is consistent and backwards compatible.)

PR #829, opened 2025-11-21
Don't reset input buffer on open (PR 250 replacement)

### LINKED ISSUE #249: Early data missed on open on Posix systems

kurtmckee:
This PR resolves merge conflicts with PR #250 so that it can be merged.

Thanks to @vlovich for this work!

Fixes #249

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -354,8 +354,6 @@ class Serial(SerialBase, PlatformSpecific):
                 if e.errno not in (errno.EINVAL, errno.ENOTTY):
                     raise
 
-            self._reset_input_buffer()
-
             self.pipe_abort_read_r, self.pipe_abort_read_w = os.pipe()
             self.pipe_abort_write_r, self.pipe_abort_write_w = os.pipe()
             fcntl.fcntl(self.pipe_abort_read_r, fcntl.F_SETFL, os.O_NONBLOCK)


dpgeorge:
@kurtmckee thanks for reworking the original PR.

Although, is it really worth adding the merge commit?  It's a very simple one-line change, made many commits ago, and I don't see the benefit in keeping the history and the merge commit.  You can see that all 30 or so recent commits have been using a rebase workflow, and it's not possible to do a rebase with this PR as it stands.

kurtmckee:
I don't share your concerns about merge commits, as I much more strongly prioritize attribution and commit signatures. However, as this is the second time that you've expressed concern about the git history, and it appears that squash merging isn't going to be used to merge these changes, I'll rework this and force-push.

kurtmckee:
@dpgeorge This is updated, and I'll use merge-less techniques moving forward. :+1:

dpgeorge (approved):

dpgeorge:
> I don't share your concerns about merge commits, as I much more strongly prioritize attribution and commit signatures. However, as this is the second time that you've expressed concern about the git history, and it appears that squash merging isn't going to be used to merge these changes, I'll rework this and force-push.

Thanks for reworking it.

There are of course different git workflows that everyone has a preference for.  Even with a merge workflow there's still the merge commit that needs to be done (usually not by the original patch author) which sometimes requires resolving conflicts, and it's not always clear how to resolve them.

Anyway, this is just my opinion, and happy to use something different for this project if others agree (but I know that @projectgus and @andrewleech prefer rebase workflow).

PR #830, opened 2025-11-21
Retry `termios.tcdrain()` on EINTR

### LINKED ISSUE #278: `termios.tcdrain` can fail with EINTR

### LINKED ISSUE #785: flush() doesn't like signals
In https://github.com/AstraLuma/klipper-pintail, i'm using `signal.setitimer()` and `SIGALRM` on a main thread and i'm reading a serial port on a daemon thread.

If the timer goes off and triggers a SIGALRM while `.flush()` is happening, I get an error like:

```
  File "/home/astraluma/pintail/pintail/systems/dwin_screen.py", line 185, in _send
    self.port.flush()
  File "/home/astraluma/pintail-env/lib/python3.9/site-packages/serial/serialposix.py", line 673, in flush
    termios.tcdrain(self.fd)
termios.error: (4, 'Interrupted system call')
```

The use of `signal.siginterrupt()` seemingly has no effect. The termios manpage provides no illumination on automatic `EINTR` or ignoring signals.

kurtmckee:
This was originally posted in PR #279, but needed to be updated to resolve merge conflicts and Python 2 syntax.

As noted in issue #278, although CPython has generally patched system calls to retry automatically on EINTR errors (see [PEP 475](https://peps.python.org/pep-0475/)), the CPython `tcdrain()` implementation doesn't appear to be catching and retrying, so this can still be an issue.

Fixes #278
Fixes #785

### DIFF
--- a/serial/serialposix.py
+++ b/serial/serialposix.py
@@ -683,7 +683,14 @@ class Serial(SerialBase, PlatformSpecific):
         """
         if not self.is_open:
             raise PortNotOpenError()
-        termios.tcdrain(self.fd)
+        while True:
+            try:
+                termios.tcdrain(self.fd)
+                break
+            except termios.error as e:
+                if e.args[0] == errno.EINTR:
+                    continue
+                raise
 
     def _reset_input_buffer(self):
         """Clear input buffer, discarding all that is in the buffer."""


dpgeorge (approved):
Thanks, this looks correct to me.  PEP 475 is definitely supposed to eliminate the need to do this in Python, and maybe this is something upstream Python would fix one day in `tcdrain`.  But for now it makes sense to handle this here.

PR #831, opened 2025-11-21
Rephrase the `read_until()` method's documentation

### LINKED ISSUE #765: Suggested rewording for read_until() docstring
Great package, thanks for all your work in maintaining it.  Minor fix below.

The current docstring says

> Read until an expected sequence is found (‘\n’ by default), the size is exceeded or until timeout occurs. If a timeout is set it may return fewer characters than requested. With no timeout it will block until the requested number of bytes is read.

It would be more precise and correct to say

> Read until the ``expected`` sequence (‘\n’ by default) has been read, or ``size`` bytes have been received, or ``timeout`` seconds have elapsed. If a timeout is set, the function may return fewer characters than requested. With no timeout the function will block until the ``expected`` sequence has been read, or ``size`` bytes have been received.  If termination is due to reading the ``expected`` sequence, that sequence is included at the end of the returned byte array or string.

kurtmckee:
This rephrase was suggested in issue 765.

Now that I know that the API documentation is being manually managed, I'll set switching to Sphinx autodoc as a future goal so documentation is managed in-code without duplication. For now, this addresses an open issue.

Closes #765

### DIFF
--- a/documentation/pyserial_api.rst
+++ b/documentation/pyserial_api.rst
@@ -164,10 +164,14 @@ Native ports
         :return: Bytes read from the port.
         :rtype: bytes
 
-        Read until an expected sequence is found ('\\n' by default), the size
-        is exceeded or until timeout occurs. If a timeout is set it may
-        return fewer characters than requested. With no timeout it will block
-        until the requested number of bytes is read.
+        Read until the *expected* sequence is found (``b'\n'`` by default),
+        or *size* bytes have been received, or the read timeout has elapsed.
+        If a timeout is set, it may return fewer characters than requested.
+        With no timeout it will block until the *expected* sequence has been read
+        or *size* bytes have been read.
+
+        If termination is due to reading the *expected* sequence,
+        that sequence is included at the end of the returned bytes.
 
         .. versionchanged:: 2.5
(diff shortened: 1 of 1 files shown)


documentation/pyserial_api.rst
```diff
@@ -164,10 +164,14 @@ Native ports
         :return: Bytes read from the port.
         :rtype: bytes
 
-        Read until an expected sequence is found ('\\n' by default), the size
-        is exceeded or until timeout occurs. If a timeout is set it may
-        return fewer characters than requested. With no timeout it will block
-        until the requested number of bytes is read.
+        Read until the expected *sequence* is found (``b'\n'`` by default),
+        or *size* bytes have been received, or ``timeout`` seconds have elapsed.
```
dpgeorge: Should "timeout" be rendered as code style?  This would be referring to the `*timeout*` argument passed to the constructor, but it's a bit unclear.

Maybe retain consistency with the rest of the docs (eg `read()` method above) and just leave it as plain text like " or the read timeout has elapsed".

kurtmckee:
_Rebased on `master` and incorporated PR feedback._

PR #832, opened 2025-11-24
Test and modernize `serial/tools/list_ports_common.py`

kurtmckee:
> [!NOTE]
>
> This PR is broken into two commits to make reviewing easier.
>
> * f1b57c65800458396ae6fa8e1458aa254dcb05d8 adds a new test file and pushes code coverage of `serial/tool/list_ports_common.py` to 100%.
>
>   It also fixes a `TypeError` crash in the `numsplit()` code, though this crash is unlikely to occur in practice.
>
> * 84cca8c28f464563c160be79f4c43a50b8469df7 modernizes the code in `serial/tool/list_ports_common.py`, primarily by adding type annotations and running the file through pyupgrade, isort, black, ruff, and mypy.
>
>   It also removes the `.apply_usb_info()` method, which was not consistently used to mutate instances and doesn't represent significant functionality.

This PR introduces the following changes:

* Test `serial.tools.list_ports_common` to 100%.

* Fix an unlikely `TypeError` that can occur
  when comparing two `numsplit()`return values.

  The crash occurs when one *text* starts with numeric characters
  and the other *text* starts with non-numeric characters,
  which results in comparisons like `[1] < ['a']`.

  The crash is unlikely because `numsplit()` is used to split device
  paths (like `"/dev/..."`) or names (like ``"COM1"` on Windows)
  that always start with non-numeric characters,
  which makes the crash unlikely based on usage.

  The crash is entirely avoided by changing the `numsplit()` return type
  from `list[str, int, ...] | list[int, str, ...]`
  to `tuple[tuple[int, ...], ...]`.

* Add pyfakefs as a test suite dependency.

  pyfakefs allows filesystem mocking so that devices and symlinks
  can "exist" for testing purposes.

* Add type annotations to `serial/tools/list_ports_common.py`.

  The code was run through mypy in strict mode and passed,
  but the annotations are not yet guaranteed and enforced
  until more code is type-annotated and mypy is added to the test suite.

* Run `serial/tools/list_ports_common.py` through multiple linters:
  pyupgrade, isort, black, ruff, and flake8.

  With the exception of double quotation marks,
  all linters are satisfied with the file now.

* Remove `if __name__ == '__main__'` code,
  which did not perform anything functionally useful.

* Remove `ListPortInfo.apply_usb_info()`,
  which is not consistently used by platform-specific port-lister code
  and does not represent significant functionality.

### DIFF
--- a/serial/tools/list_ports_common.py
+++ b/serial/tools/list_ports_common.py
@@ -8,20 +8,20 @@
 #
 # SPDX-License-Identifier:    BSD-3-Clause
 
-from __future__ import absolute_import
+from __future__ import annotations
 
-import re
 import glob
-import os
 import os.path
+import re
+import typing
 
 
-def numsplit(text):
+def numsplit(text: str) -> tuple[tuple[int, ...], ...]:
     """\
     Convert string into a list of texts and numbers in order to support a
     natural sorting.
     """
-    result = []
+    result: list[tuple[int, ...]] = []
     for group in re.split(r'(\d+)', text):
         if group:
             try:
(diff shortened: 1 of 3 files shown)


kurtmckee:
@zsquareplusc @andrewleech Please review this PR when you have an opportunity. Future testing work is dependent on having pyfakefs available, and I'm blocked on this merging so I can avoid merge conflicts.

PR #838, opened 2025-12-08
Begin testing sysfs port recognition

kurtmckee:
> [!NOTE]
>
> This PR is broken into two commits, and should likely be reviewed one at-a-time.
>
> * f5cb780 introduces a new command, `pyserial-generate-test-asset`, which can generate a JSON output of a user's sysfs-based device information, and introduces a test framework for recreating the device's sysfs filesystem entry in-memory as it existed on the user's system.
>
>   This will help with debugging sysfs-based device recognition, as well as locking valuable information into the test suite to help evaluate future code changes.
>
>   Three devices from my own testing, representing the `usb`, `usb-serial`, and `serial-base` sysfs subsystems, are now tested.
>
> * 323c81f updates/modernizes the code in `list_ports_linux.py`, which is where the sysfs-based device recognition code lives. The code was run through pyupgrade, isort, and has had type annotations added. A cursory run of mypy was made and some issues were resolved, but type annotations are not guaranteed to be correct until more of the code has been type-annotated.

* Add a `pytest-generate-test-asset` command.

  The command produces a detailed JSON output
  to capture the state of a device on a system
  that can be used to diagnose issues with device recognition
  and can be added directly to the test suite for permanence.

  This command only works on sysfs systems so far.

* Add test assets generated from an Ubuntu 24.04 system
  running kernel 6.14.0 with real devices plugged in.

  These assets demonstrate the following subsystems:

  - `serial-base`
  - `usb`
  - `usb-serial`

  and demonstrate globbing of the following `/dev` patterns:

  - `/dev/ttyACM*`
  - `/dev/ttyS*`
  - `/dev/ttyUSB*`

* Increase the code coverage `fail-under` value to 30%.
  Previously the value was 20%.

* mypy was run against the new CLI command and passes.

  However, type annotations are not yet guaranteed
  until more of the project is type annotated
  and mypy is passing in CI.

* Type annotations have been added to `list_ports_linux.py`.
* The `list_ports_linux.py` code has been run through mypy.
  Some issues were found and fixed,
  but others will require additional work.
* The code has been run through pyupgrade and isort.
* Device patterns (like `/dev/ttyS*`) have been moved,
  and have been reordered to reflect which patterns have tests
  and which are currently untested.
* The `if __name__ == "__main__"` condition has been removed.
  If seeing the device subsystem is valuable,
  the `pyserial-ports` command can be updated for that purpose.
* Some typos were fixed.


serial/tools/list_ports_linux.py
```diff
@@ -8,28 +8,29 @@
 #
 # SPDX-License-Identifier:    BSD-3-Clause
```
andrewleech: New files should have a SPDX header too

serial/tools/list_ports_linux.py
```diff
@@ -61,54 +62,60 @@ def __init__(self, device):
...
         devices.update(list_ports_common.list_links(devices))
-    return [info
-            for info in [SysFS(d) for d in devices]
-            if info.subsystem != "platform"]    # hide non-present internal serial ports
-
-# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-# test
-if __name__ == '__main__':
-    for info in sorted(comports()):
-        print("{0}: {0.subsystem}".format(info))
```
andrewleech: Was it neccesary to remove this?  As much as formal testing is done with pytest I often like little test harnesses like this myself.

andrewleech (commented):
This generated mocking does feel like a fairly pragmatic way to get closer-to-real-world test coverage. 

It think it really needs a bit more documentation than the comment looking for more sample devices, it took me a bit to figure out how it's all fitting together.

I'm also not sure if it really needs the generate function exposed as a cli tool, surely a documented `python -c 'import serial; serial.tools.generate...` sort of line is more appropriate for the small number of people who might run it once-off to capture a device? Then some more description of how the test itself should be written to use the new asset.

serial/tools/generate_test_asset/_sysfs.py
```diff
@@ -0,0 +1,128 @@
...
+            raw_contents = file.read(1001)
+    except OSError as error:
+        raw_contents = b""
+        os_error: tuple[typing.Any, ...] | None = error.args
+    else:
+        os_error = None
+
+    # Truncate if necessary.
+    is_truncated = False
+    if len(raw_contents) > 1000:
```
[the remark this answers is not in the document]
kurtmckee: I haven't seen this. It's exclusively for cases that I haven't seen, so that it's clearly marked for developers reviewing an incoming issue that such-and-such file was much larger than expected.

serial/tools/list_ports_linux.py
kurtmckee: It's not necessary for this PR. However, these little chunks of untestable code need to be removed.

The existing `pyserial-ports` command is where this should live, which does all of the same work using tested code. It just needs to be extended to include the subsystem in its output.

kurtmckee:
> I'm also not sure if it really needs the generate function exposed as a cli tool

It might not need to be permanent, but the volume of issues and PRs associated with incorrect port detection is something I want to whittle down, and having a namespaced `pyserial-*` command is a low-friction request of users.

Tucking this into a `python -c` invocation (a) increases friction, and (b) introduces chances of invocation failures due to incompatibilities between platforms' and shells' quoting rules (Windows' `cmd.exe` being a prime example that I've handled support tickets for in the last year).

I'd prefer to keep the command as-is; I believe it has no cost, not even to tab completion.

> it really needs a bit more documentation

I agree, and I'm very open to suggestions for what else to add here!

> some more description of how the test itself should be written to use the new asset

I'll try to figure out the best place to add that and update the PR!

kurtmckee:
@andrewleech I've added a `CONTRIBUTING.rst` page, which will show up as a tab in the GitHub repo view:

> image

I've added that to the built documentation as well.

I've also added a `README` to the `test/assets/` directory to help identify how files are added, and how new tests are constructed. I think this addresses the documentation suggestion.

kurtmckee:
_Rebased on `master`, resolved merge conflicts, and force-pushed._

In addition to resolving the merge conflicts, the test assets were regenerated to include the `/proc/tty/drivers` file that is now needed. I also modified the test file generator so that it splits file contents by newlines so that the contents serialize to JSON as lists of strings. This allows for much easier reading of the file contents when reviewing such files in the future.

```diff
- "contents": "/dev/tty             /dev/tty        5       0 system:/dev/tty\n/dev/console         /dev/console    5       1 system:console",
+ "contents": [
+   "/dev/tty             /dev/tty        5       0 system:/dev/tty",
+   "/dev/console         /dev/console    5       1 system:console",
+ ],
```

PR #839, opened 2025-12-13
Remove `time.sleep()` calls from `test_threaded.py`

kurtmckee:
This change reduces the test suite runtime from ~11 seconds to ~9 seconds by replacing two calls to `time.sleep()` in `test_threaded.py` with a queue that blocks only exactly the amount of time required for a reader thread to put its data to the queue.

In a follow-up commit, `test_threaded.py` was run through a number of linters to resolve code quality and style issues.

### DIFF
--- a/test/test_threaded.py
+++ b/test/test_threaded.py
@@ -8,17 +8,17 @@
 Test serial.threaded related functionality.
 """
 
-import os
 import queue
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
A nice improvement, thanks!

PR #840, opened 2025-12-13
Fixup `test_readline.py`

kurtmckee:

* Remove Python 2 compatibility code
* Use bytes literals for clarity
* Use pytest, not unittest, for simplicity
* Remove a timeout in the test code, which reduces test suite runtimes by ~3 seconds

### DIFF
--- a/test/test_readline.py
+++ b/test/test_readline.py
@@ -22,20 +22,12 @@ On a 9 pole DSUB these are the pins (2-3) (4-6) (7-8)
 """
 
 import unittest
-import sys
 import serial
 
 #~ print serial.VERSION
 
 # on which port should the tests be performed:
 PORT = 'loop://'
-
-if sys.version_info >= (3, 0):
-    def data(string):
-        return bytes(string, 'latin1')
-else:
-    def data(string):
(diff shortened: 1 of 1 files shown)


test/test_readline.py
```diff
@@ -4,101 +4,39 @@
...
-Intended to be run on different platforms, to ensure portability of
-the code.
+import typing
 
-For all these tests a simple hardware is required.
-Loopback HW adapter:
-Shortcut these pin pairs:
- TX  <-> RX
- RTS <-> CTS
- DTR <-> DSR
```
dpgeorge: It feels this comment, and the `if __name__ == '__main__':` code below, are useful because they allow testing on hardware.  I know that's something @zsquareplusc does before a release (test on various serial hardware).  So maybe they should be kept?

kurtmckee: All of these little scripts, one-offs, and entry points need to completely disappear in favor of a comprehensive entrypoint that executes a known codepath by which real hardware gets tested.

That said, if @zsquareplusc states that this code path is needed in the interim then I can add it back in.

PR #842, opened 2025-12-14
Thoroughly test the `Timeout` class

kurtmckee:
This introduces the following changes to the `Timeout` class:

* Eliminate `TIME` and just use `time.monotonic()`.
* Eliminate conditionals that check for time having jumped. (`time.monotonic()` guarantees time always moves forward.)
* Add type annotations and fix issues reported by mypy.

Also, these changes were introduced to `test_timeout_class.py`:

* Remove `time.sleep()` calls in favor of mocking. This reduces the test suite runtime by ~2.1 seconds.
* Use pytest constructs, not unittest.

This PR supersedes #231.

### DIFF
--- a/serial/serialutil.py
+++ b/serial/serialutil.py
@@ -103,63 +103,53 @@ class PortNotOpenError(SerialException):
         super(PortNotOpenError, self).__init__('Attempting to use a port that is not open')
 
 
-class Timeout(object):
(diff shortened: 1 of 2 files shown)


dpgeorge (approved):
Another nice clean up, thank you!

PR #846, opened 2025-12-16
Update the CHANGELOG

kurtmckee:
This documents all PRs and commits since pyserial 3.5 was released, and introduces sphinx-issues so that PRs, issues, and commits can be easily linked to.

The links render like this:

> image

It will be helpful to switch from this style of manual changelog management to using tools for this purpose.

I recommend [scriv](https://scriv.readthedocs.io/en/latest/) so that PRs can document themselves with narrative text without worrying about merge conflicts caused by editing the same file during development. I can introduce a PR for this if desired.


kurtmckee:
_Rebased on `master`, resolved merge conflicts, and force-pushed._

PR #847, opened 2025-12-22
Fix USB interface detection on macOS by using modern IOUSBHostInterfa…

BBBiram:
…ce API

- Add support for IOUSBHostInterface (modern) with fallback to IOUSBInterface (legacy)
- Replace scan_interfaces() and search_for_locationID_in_interfaces() with direct lookup
- Add get_interface_name_from_service() helper function for cleaner interface name retrieval
- Use kUSBString property instead of deprecated 'USB Interface Name'
- Remove intermediate SuitableSerialInterface class that is no longer needed

This fixes interface name detection on newer macOS versions where IOUSBHostInterface replaced the deprecated IOUSBInterface, and simplifies the code by eliminating the two-pass scan approach.

### DIFF
--- a/serial/tools/list_ports_osx.py
+++ b/serial/tools/list_ports_osx.py
@@ -232,30 +232,13 @@ class SuitableSerialInterface(object):
     pass
 
 
-def scan_interfaces():
-    """
-    helper function to scan USB interfaces
-    returns a list of SuitableSerialInterface objects with name and id attributes
-    """
-    interfaces = []
-    for service in GetIOServicesByType('IOSerialBSDClient'):
-        device = get_string_property(service, "IOCalloutDevice")
-        if device:
-            usb_device = GetParentDeviceByType(service, "IOUSBInterface")
-            if usb_device:
-                name = get_string_property(usb_device, "USB Interface Name") or None
-                locationID = get_int_property(usb_device, "locationID", kCFNumberSInt32Type) or ''
-                i = SuitableSerialInterface()
-                i.id = locationID
-                i.name = name
-                interfaces.append(i)
-    return interfaces
-
-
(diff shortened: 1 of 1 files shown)


alonbl (approved):

harryberlin:
does ist Work for not modern Version?

BBBiram:
Yes. Although this was tested on Apple Silicon, it also applies to older (non-modern) macOS versions and Intel Macs.
Apple redesigned the USB stack in OS X 10.11 (El Capitan, 2015) and introduced IOUSBHostFamily as the replacement for the legacy IOUSBFamily.
Since then, IOUSBInterface has been deprecated, and Apple recommends using IOUSBHostInterface for USB drivers. This change applies regardless of CPU architecture.

Sources:
[1] [https://developer.apple.com/support/kernel-extensions/](https://developer.apple.com/support/kernel-extensions/)
[2] [https://developer.apple.com/documentation/kernel/iousbinterface](https://developer.apple.com/documentation/kernel/iousbinterface)

projectgus (approved):
I don't have a Mac to test this on, but this change looks reasonable to me - thanks @BBBiram!

dpgeorge:
I have an old Intel MacBook Air from 2013 running Big Sir (11.7.10).  I tested this PR and the same string is returned for `.interface` before and after this PR (it returns `USB_CDC1`).

dpgeorge (approved):

BBBiram:
Just a friendly follow-up. This PR has been approved for quite some time, so I wanted to check whether there’s anything else needed from my side before it can be merged.

If everything looks good, I’d really appreciate it if one of the maintainers could merge it when convenient. If there are any remaining concerns or changes you’d like me to make, I’m happy to update the PR.

Thank you for your time and for maintaining pySerial!

PR #861, opened 2026-04-28
FreeBSD Support for list_ports

kev009:
Fixes over the original PR https://github.com/pyserial/pyserial/pull/778:

- Call super().__init__() so base ListPortInfo attributes are initialized
- Parse angle-bracket description properly with regex instead of split() which mangled the USB description into props
- Set product to the USB description string instead of setting manufacturer to the integer VID
- Strip surrounding quotes from serial numbers
- Handle include_links in comports()
- Add subprocess timeout and error handling

### DIFF
--- a/serial/tools/list_ports_freebsd.py
+++ b/serial/tools/list_ports_freebsd.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python
+#
+# This is a module that gathers a list of serial ports including details on
+# FreeBSD systems, using devinfo(8) to discover USB device information.
+#
+# This file is part of pySerial. https://github.com/pyserial/pyserial
+# (C) 2011-2015 Chris Liechti <cliechti@gmx.net>
+# FreeBSD support by:
+#  Poul-Henning Kamp <phk@FreeBSD.org> and
+#  Kevin Bowling <kbowling@FreeBSD.org>
+# SPDX-License-Identifier:    BSD-3-Clause
+
+from __future__ import absolute_import
+
+import glob
+import re
+import subprocess
+
+from serial.tools import list_ports_common
+
+
+class DevInfo(list_ports_common.ListPortInfo):
+    """Collect serial port info from FreeBSD devinfo(8) output."""
+
+    def __init__(self, device, props=None, usb_desc=None):
+        super(DevInfo, self).__init__(device, skip_link_detection=True)
+        self.props = props or {}
+
+        if "vendor" in self.props:
+            self.vid = int(self.props["vendor"], 16)
+        if "product" in self.props:
+            self.pid = int(self.props["product"], 16)
+        if "sernum" in self.props:
+            self.serial_number = self.props["sernum"].strip('"')
+
+        # The USB description from devinfo contains the combined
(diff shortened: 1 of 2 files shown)


sbz:
Thanks @kev009! Looks strong :muscle:  

Maybe you should use the [Co-authored by](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-github) tag in your commit to also mention effort of @bsdphk

serial/tools/list_ports_freebsd.py
```diff
@@ -0,0 +1,129 @@
...
+            addr 2> pnpinfo vendor=0x1209 product=0x4d69 ... ttyname=U0 \
+            ... ugen=ugen1.3
+
+    Returns (device_path, props_dict, usb_description) or None.
+    """
+    props = {}
+    usb_desc = None
+
+    # Match the full devinfo format with angle-bracket USB description
+    m = re.match(r'\s*\S+\s+<([^>]*)>.*?pnpinfo\s+(.*)', line)
```
sbz:
```suggestion
    m = re.match(r'\s*\S+\s+<(?P<desc_text>[^>]*)>.*?pnpinfo\s+(?P<props_text>.*)', line)
```

Just a style change, using regex, I would suggest to use [named group](https://docs.python.org/3/howto/regex.html#non-capturing-and-named-groups) to be explicit. 

And indeed what you called the remainder is the `props` attributes cf. [print_device_props](https://github.com/freebsd/freebsd-src/blob/main/usr.sbin/devinfo/devinfo.c#L159) in devinfo(8), I think it should be renamed from `remainder` to `props_text`

serial/tools/list_ports_freebsd.py
```diff
@@ -0,0 +1,129 @@
...
+    if m:
+        desc_text = m.group(1)
```
sbz:
```suggestion
        desc_text = m.group('desc_text')
```

serial/tools/list_ports_freebsd.py
```diff
@@ -0,0 +1,129 @@
...
+        # Extract the human-readable part before ", class "
+        desc_parts = desc_text.split(', class ')
+        if desc_parts:
+            usb_desc = desc_parts[0].strip()
+        remainder = m.group(2)
```
sbz:
```suggestion
        props_text = m.group('props_text')
```

serial/tools/list_ports_freebsd.py
```diff
@@ -0,0 +1,129 @@
...
+    else:
+        remainder = line
```
sbz:
```suggestion
        props_text = line
```

serial/tools/list_ports_freebsd.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # Parse key=value pairs from pnpinfo or fallback line
+    for token in remainder.split():
```
sbz:
```suggestion
    for token in props_text.split():
```

sbz (commented):
Minor style feedback

PR #868, opened 2026-07-12
Fix bare except clauses in rfc2217.py and miniterm.py

koteshyelamati:
Fix 3 bare `except:` clauses that catch `BaseException` (including `SystemExit`, `KeyboardInterrupt`), replacing them with specific exception types.

**`serial/rfc2217.py`:**
- Line 493: `except:` → `except Exception:` — wraps port open/configure operations; cleans up and re-raises on any error.
- Line 545: `except:` → `except OSError:` — wraps `socket.shutdown()` and `socket.close()` which raise `OSError` on failure; comment says "ignore errors."

**`serial/tools/miniterm.py`:**
- Line 557: `except:` → `except Exception:` — wraps the main writer loop; sets `self.alive = False` and re-raises on error.

Bare `except:` clauses are bad practice (PEP 8 / flake8 E722) because they silently swallow `KeyboardInterrupt` and `SystemExit`, which can prevent clean program shutdown.

### DIFF
--- a/serial/rfc2217.py
+++ b/serial/rfc2217.py
@@ -490,7 +490,7 @@ class Serial(SerialBase):
                 self._update_rts_state()
             self.reset_input_buffer()
             self.reset_output_buffer()
-        except:
+        except Exception:
             self.close()
             raise
 
@@ -542,7 +542,7 @@ class Serial(SerialBase):
             try:
                 self._socket.shutdown(socket.SHUT_RDWR)
                 self._socket.close()
-            except:
+            except OSError:
                 # ignore errors.
                 pass
         if self._thread:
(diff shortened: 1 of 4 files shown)


fallenmi (changes requested):
The RFC 2217 change removes cleanup when an interrupt arrives after the socket and reader thread have been allocated.

An exact base/head probe ran `Serial.open()` at base `a5c48d4` and head `0f2b0f7` with a fake socket/thread and `_reconfigure_port()` raising `KeyboardInterrupt("probe")`. The base re-raised only after `close()` left `is_open == False`, shut and closed the socket, cleared `_socket`, joined and cleared `_thread`. The head re-raised with `is_open == True`, the socket still open and retained, and the thread still retained.

This block does not swallow `KeyboardInterrupt` or `SystemExit`: the bare `raise` already propagates them after resource cleanup. Please keep cleanup for all `BaseException` paths here (for example, `except BaseException: self.close(); raise`) while narrowing only the genuinely ignored exception handlers, and add a regression for an interrupt after allocation. The existing `test/test_rfc2217.py` target passes on the reviewed head; compile and diff checks are also clean.

Reviewed with OpenAI Codex assistance.

PR #873, opened 2026-08-14
Keep list_ports importable on Android

gszep:
## Summary
- add an Android/Termux branch to `serial.tools.list_ports_posix`
- scan for `/dev/ttyACM*` and `/dev/ttyUSB*` when such nodes are available
- return an empty list instead of raising ImportError on typical unrooted Android systems

## Context
On Termux, `sys.platform` is `android`. Importing `serial.tools.list_ports` currently raises from `list_ports_posix.py`, which can break applications before they have a chance to open an explicit port or custom URL.

This does not implement Android USB host support; Android still often exposes USB devices only through the Android USB APIs/Termux USB file descriptors. It simply makes `list_ports` behave like a valid-but-empty enumerator unless tty device nodes exist.

## Testing
- `python -m py_compile serial/tools/list_ports_posix.py`
- Hardware context: reproduced on Android/Termux with a USB CDC ACM ESP32-S2/S3-class device where no `/dev/ttyACM*` or `/dev/ttyUSB*` nodes were exposed.

### DIFF
--- a/serial/tools/list_ports_posix.py
+++ b/serial/tools/list_ports_posix.py
@@ -104,6 +104,20 @@ elif plat[:5] == 'haiku':    # Haiku
             devices.update(list_ports_common.list_links(devices))
         return [list_ports_common.ListPortInfo(d) for d in devices]
 
+elif plat[:7] == 'android':   # Android / Termux
+    def comports(include_links=False):
+        """scan for serial ports exposed as device nodes.
+
+        Most unrooted Android/Termux environments do not expose USB serial
+        adapters as /dev/ttyACM* or /dev/ttyUSB*. Returning an empty list keeps
+        serial.tools.list_ports importable; applications that know a concrete
+        device path can still try to open it explicitly.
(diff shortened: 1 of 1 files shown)


gszep:
Follow-up context from the Android/Termux hardware test:

The small `list_ports` Android change here is not sufficient by itself for USB-OTG flashing, because unrooted Android still exposes USB devices via Android USB APIs/Termux file descriptors rather than POSIX tty nodes.

However, keeping `serial.tools.list_ports` importable remains useful: in the same test setup, a separate Termux USB CDC bridge exposed the Android USB device to a workstation over pySerial-compatible transports. pySerial's RFC2217 client/server pieces could then be used for normal serial traffic and control-line experiments without crashing during imports on `sys.platform == 'android'`.

So I still see this PR as a narrow compatibility improvement: Android may enumerate zero tty ports, but importing list_ports should not prevent applications from using explicit paths, URL handlers, or external bridges.

